PR #148, opened 2015-02-05
Add a new "Polls" example

kylef:
Here's the start of the new example API. @zdne please take a look at this, feedback is very welcome!

### DIFF
--- a/examples/Polls API.md
+++ b/examples/Polls API.md
@@ -2,7 +2,7 @@ FORMAT: 1A
 
 # Polls
 
-Polls is a simple web service that allows consumers to view polls and vote in them.
+Polls is a simple web service that allows consumers to view polls and vote in them. You can view this documentation over at [Apiary](http://docs.pollsapi.apiary.io).
 
 # Polls API Root [/]
 
@@ -24,12 +24,12 @@ Resource related to questions in the API.
 
 ## Question [/questions/{question_id}]
 
-A Question object has the following attributes.
+A Question object has the following attributes:
 
 - question
-- published_at
+- published_at - An ISO8601 date when the question was published.
 - url
-- choices (an array of Choice objects).
+- choices - An array of Choice objects.
 
 + Parameters
 
--- a/examples/Polls API.md
+++ b/examples/Polls API.md
@@ -1,0 +1,179 @@
+FORMAT: 1A
+
+# Polls
+
+Polls is a simple web service that allows consumers to view polls and vote in them.
+
+# Polls API Root [/]
+
+This resource does not have any attributes. Instead it offers the initial API affordances in the form of the links in the JSON body.
+
+It is recommend to follow the “url” link values or Link headers to get to resources instead of constructing your own URLs to keep your client decoupled from implementation details.
+
+## Retrieve the Entry Point [GET]
+
++ Response 200 (application/json)
+
+        {
+            "questions_url": "/questions{?page}"
+        }
+
+## Group Question
+
+Resource related to questions in the API.
+
+## Question [/questions/{question_id}]
+
+A Question object has the following attributes.
+
+- question
+- published_at
+- url
+- choices (an array of Choice objects).
+
++ Parameters
+
+    + question_id (number) ... ID of the Question in form of an integer
+
+### View a question detail [GET]
+
++ Response 200 (application/json)
+
+                {
+                    "question": "Favourite programming language?",
+                    "published_at": "2014-11-11T08:40:51.620Z",
+                    "url": "/questions/1",
+                    "choices": [
+                        {
+                            "answer": "Swift",
+                            "url": "/questions/1/choices/1",
+                            "votes": 2048
+                        }, {
+                            "answer": "Python",
+                            "url": "/questions/1/choices/2",
+                            "votes": 1024
+                        }, {
+                            "answer": "Objective-C",
+                            "url": "/questions/1/choices/3",
+                            "votes": 512
+                        }, {
+                            "answer": "Ruby",
+                            "url": "/questions/1/choices/4",
+                            "votes": 256
+                        }
+                    ]
+                }
+
+## Choice [/questions/{question_id}/choices/{choice_id}]
+
++ Parameters
+
+    + question_id (number) ... ID of the Question in form of an integer
+    + choice_id (number) ... ID of the Choice in form of an integer
+
+### Vote on a Choice [PUT]
+
+This action allows you to vote on a question's choice.
+
++ Response 201
+
+    + Headers
+
+            Location: /questions/1
+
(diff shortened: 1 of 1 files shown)


kylef:
Forgot to mention, you can find an example of this rendered at Apiary over at http://docs.pollstutorialapi.apiary.io/

examples/Polls API.md
```diff
@@ -0,0 +1,179 @@
...
+
+# Polls
+
+Polls is a simple web service that allows consumers to view polls and vote in them.
+
+# Polls API Root [/]
+
+This resource does not have any attributes. Instead it offers the initial API affordances in the form of the links in the JSON body.
+
+It is recommend to follow the “url” link values or Link headers to get to resources instead of constructing your own URLs to keep your client decoupled from implementation details.
```
zdne: I like this sentence! Except the blueprint doesn't show the Link HTTP header so maybe throw that mention out?

examples/Polls API.md
```diff
@@ -0,0 +1,179 @@
...
++ Parameters
+
+    + question_id (number) ... ID of the Question in form of an integer
+
+### View a question detail [GET]
+
++ Response 200 (application/json)
+
+                {
+                    "question": "Favourite programming language?",
```
zdne: Oh my its in Traditional English!  :gb:

examples/Polls API.md
```diff
@@ -0,0 +1,179 @@
...
+                        }
+                    ]
+                }
+            ]
+
+### Create a new question [POST]
+
+You can create your own question using this action. It takes a JSON dictionary containing a question and a collection of answers in the form of choices.
+
+- question (string) - The question
```
zdne: just a nitpick: you are mixing styles and using both `-` and `+` for lists

examples/Polls API.md
```diff
@@ -0,0 +1,179 @@
...
+
+    + page (optional, number) ... The page of questions to return
+
+### List all questions [GET]
+
++ Response 200 (application/json)
+
+    + Headers
+    
+            Link: </questions?page=2>; rel="next"
```
zdne: Oh so there is a Link header. Hmm I think it may be a little confusing to mention it in the beginning of this document when its first appearance is here. Not sure what would be a better way tho. I really like this style / tone you are using there :+1:

zdne:
@kylef reviewed

examples/Polls API.md
honzajavorek: So colourful! :+1:

zdne:
> Forgot to mention, you can find an example of this rendered at Apiary over at http://docs.pollstutorialapi.apiary.io/

I think it would be good to mention in on the top of the blueprint?

zdne:
@kylef poke – this perhaps needs some updates (see my review), or do you think it is fine as it is?

examples/Polls API.md
```diff
@@ -0,0 +1,179 @@
...
+            "questions_url": "/questions{?page}"
+        }
+
+## Group Question
+
+Resource related to questions in the API.
+
+## Question [/questions/{question_id}]
+
+A Question object has the following attributes.
```
zdne: Colon – `attributes:` ?

examples/Polls API.md
```diff
@@ -0,0 +1,179 @@
...
+
+- question
+- published_at
+- url
+- choices (an array of Choice objects).
```
zdne: I would say use the MSON syntax even it it is not in the scope of this blueprint to explain it e.g. 

```
- choices - an array of Choice objects
```

It will be more consistent in the grand picture later

examples/Polls API.md
```diff
@@ -0,0 +1,179 @@
...
+                    "Objective-C",
+                    "Ruby"
+                ]
+            }
+
++ Response 201
+
+    + Headers
+
+            Location: /questions/1
```
zdne: So you are using, correctly, a "new" header here. Previously you were recommending following the `Link` header and / or `_url` properties. Shouldn't those recommendation include the `Location` header as well. 

Bottom line: I think it will feel confusing and/or overwhelming for a newcomer to have 3 different points where he can get an URL to follow..

zdne:
@kylef added some new remarks please address them before I will merge

examples/Polls API.md
kylef: To address your comments about it being later and not properly portrayed here. Would the following sentence be any better?

> It is recommend to follow the “url” link values or [Link headers](https://tools.ietf.org/html/rfc5988) where applicable to get to resources instead of constructing your own URLs to keep your client decoupled from implementation detail

Adding "where applicable" to hint that it's not always there?

examples/Polls API.md
kylef: I didn't think it was worth explicitly mentioning this one because it's part of HTTP [specification](http://tools.ietf.org/html/rfc2616). Everything in this spec should be expected.

examples/Polls API.md
kylef: Perhaps instead the following, which mentions location headers. But this is perhaps too much information.

> Rather than constructing your own URLs, it is recommended that you following links exposed in the API instead to keep your client decoupled from implementation details. These links will be exposed in `url` link values in response bodies, along with [`Link`](https://tools.ietf.org/html/rfc5988) or [`Location`](http://en.wikipedia.org/wiki/HTTP_location) headers.

zdne: Yeah, it might be too much information. But how about just 

> It is recommend to follow the “url” link values or Link or Location headers where applicable ...

?

examples/Polls API.md
kylef: The other tutorials such as Gist Fox also are doing this. Personally I have been using the `+` for API Blueprint sections such as `Response` `Parameters` etc. But used `-` for other lists.

I'm happy to change these to `+` along with the other tutorials if you would rather them all use `+`.

kylef:
@zdne, I've addressed some of the feedback. Please take another look at this one.

zdne:
Hmm maybe I would throw out the 

> Again, instead of constructing the URLs for the next page. It is highly recommended that you follow the next link header in the response.

Paragraph completely. It feels like we are way too zealous with it? What do you think @kylef ?

kylef:
@zdne I think you're right, this kind of message doesn't really fit in a specific resources description but instead the introduction section. I've pulled it out.

examples/Polls API.md
zdne: I am fine with either. But maybe being consistent with list will be easier for a first-time reader, so he doesn't have to spent any time thinking why there are sometimes `-`s and sometimes `+`s ? This is a minor issue tho.

zdne: Speaking of which we really need to revisit our older examples and simplify / clarify them...

PR #149, opened 2015-02-06
Put HOST after FORMAT

BigBlueHat:
The @apiaryio editor complained when these were switched.

Which makes sense as the first thing a parser would need to know is the format, then whatever is correct for that format.

Thanks!


zdne:
@BigBlueHat thanks sir!

PR #156, opened 2015-02-20
[Polls Example] Improve consistency

kylef:
While implementing this I noticed a few inconsistencies etc. This pull request addresses them and also provides example values for parameters so dredd can use it.

I've also filled in the HOST metadata to point to https://pollsapi.herokuapp.com/.

### DIFF
--- a/examples/Polls API.md
+++ b/examples/Polls API.md
@@ -1,4 +1,5 @@
 FORMAT: 1A
+HOST: https://pollsapi.herokuapp.com/
 
 # Polls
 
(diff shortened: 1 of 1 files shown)


fosrias:
@kylef So, this API has no endpoint to see the results of the voting on a
particular question like question/id/votes or something?

pksunkara:
@fosrias Retrieving the question will also include the votes for each choices in it.

fosrias:
@pksunkara @kylef Whenever you start embedding some other functionality like that in a payload it is begging for separate resources that you can access independently, which is my one nagging feeling. That is the resource can carry links to other resources. If you want them embedded having a parameter requesting that is a cool option. Just some thoughts.

PR #158, opened 2015-02-24
Align parameters syntax with attributes

pksunkara:
For @zdne to review & merge.

I had all this description about how the old syntax was changed to the new syntax and stuff. Then, I realized that it is not the way we do things in the API Blueprint specification. :neutral_face: 

We only describe the current syntax, but still offer backward compatibility.

### DIFF
--- a/API Blueprint Specification.md
+++ b/API Blueprint Specification.md
@@ -766,7 +766,7 @@ Defined by the `Parameters` keyword written in a Markdown list item:
 #### Description
 Discussion of URI parameters in the _scope of the parent section_.
 
-This section **must** be composed of nested list items only. This section **must not** contain any other elements. Each of the list items describes a single URI parameter. The nested list items subsections inherit from the [Named section](#def-named-section) and are subject to additional formatting as follows:
+This section **must** be composed of nested list items only. This section **must not** contain any other elements. Each list item describes a single URI parameter. The nested list items subsections inherit from the [Named section](#def-named-section) and are subject to additional formatting as follows:
 
     + <parameter name>: `<example value>` (<type> | enum[<type>], required | optional) - <description>
 
@@ -788,9 +788,9 @@ Where:
 * `<default value>` is an **optional** default value of the parameter – a value that is used when no value is explicitly set (optional parameters only).
 * `<example value>` is an **optional** example value of the parameter (e.g. `1234`).
 * `<type>` is the **optional** parameter type as expected by the API (e.g. "number", "string", "boolean"). "string" is the **default**.
-* `Members` is the **optional** enumeration of possible values. `<type>` should be surrounded by `enum[]` if this is present.
+* `Members` is the **optional** enumeration of possible values. `<type>` should be surrounded by `enum[]` if this is present. For example, if enumeration values are present for a parameter whose type is `number`, then `enum[number]` should be used instead of `number` to.
 * `<enumeration value n>` represents an element of enumeration type.
-* `<enumeration description n>` represents the description of the corresponding enumeration element.
+* `<enumeration description n>` is the **optional** description of the corresponding enumeration element.
 * `required` is the **optional** specifier of a required parameter (this is the **default**)
 * `optional` is the **optional** specifier of an optional parameter.
 
@@ -830,9 +830,9 @@ Where:
         Id of a Post
 
         + Members
-            + `A` - This is alphabet A
+            + `A` - This is character A
             + `B`
-            + `C` - This is alphabet C
+            + `C` - This is character C
 ```
 ---
 
--- a/API Blueprint Specification.md
+++ b/API Blueprint Specification.md
@@ -766,17 +766,19 @@ Defined by the `Parameters` keyword written in a Markdown list item:
 #### Description
 Discussion of URI parameters in the _scope of the parent section_.
 
-This section **must** be composed of nested list items only. This section **must not** contain any other elements. One list item per URI parameter. The nested list items subsections inherit from the [Named section](#def-named-section) and are subject to additional formatting as follows:
+This section **must** be composed of nested list items only. This section **must not** contain any other elements. Each of the list items describes a single URI parameter. The nested list items subsections inherit from the [Named section](#def-named-section) and are subject to additional formatting as follows:
 
-    + <parameter name> = `<default value>` (required | optional , <type>, `<example value>`) ... <description>
+    + <parameter name>: `<example value>` (<type> | enum[<type>], required | optional) - <description>
 
         <additional description>
 
-        + Values
-            + `<enumeration element 1>`
-            + `<enumeration element 2>`
+        + Default: `<default value>`
+
+        + Members
+            + `<enumeration value 1>` - <enumeration description 1>
+            + `<enumeration value 2>` - <enumeration description 2>
             ...
-            + `<enumeration element N>`
+            + `<enumeration value N>` - <enumeration description N>
 
 Where:
 
@@ -785,9 +787,10 @@ Where:
 * `<additional description>` is any additional **optional** Markdown-formatted [description](#SectionDescription) of the parameter.
 * `<default value>` is an **optional** default value of the parameter – a value that is used when no value is explicitly set (optional parameters only).
 * `<example value>` is an **optional** example value of the parameter (e.g. `1234`).
-* `<type>` is the **optional** parameter type as expected by the API (e.g. "number").
-* `Values` is the **optional** enumeration of possible values
-*  and `<enumeration element n>` represents an element of enumeration type.
+* `<type>` is the **optional** parameter type as expected by the API (e.g. "number", "string", "boolean"). "string" is the **default**.
+* `Members` is the **optional** enumeration of possible values. `<type>` should be surrounded by `enum[]` if this is present.
+* `<enumeration value n>` represents an element of enumeration type.
+* `<enumeration description n>` represents the description of the corresponding enumeration element.
 * `required` is the **optional** specifier of a required parameter (this is the **default**)
 * `optional` is the **optional** specifier of an optional parameter.
 
@@ -801,34 +804,35 @@ Where:
 
 ```
 + Parameters
-    + id ... Id of a post.
+    + id - Id of a post.
 ```
 
 ```
 + Parameters
-    + id (number) ... Id of a post.
+    + id (number) - Id of a post.
 ```
 
 ```
 + Parameters
-    + id (required, number, `1001`) ... Id of a post.
+    + id: `1001` (number, required) - Id of a post.
 ```
 
 ```
 + Parameters
-    + id = `20` (optional, number, `1001`) ... Id of a post.
+    + id: `1001` (number, optional) - Id of a post.
+        + Default: `20`
 ```
 
 ```
 + Parameters
-    + id (string)
+    + id (enum[string])
 
         Id of a Post
 
-        + Values
-            + `A`
+        + Members
+            + `A` - This is alphabet A
             + `B`
-            + `C`
+            + `C` - This is alphabet C
 ```
 ---
 
--- a/examples/09. Advanced Attributes.md
+++ b/examples/09. Advanced Attributes.md
@@ -20,7 +20,7 @@ The `Create a Coupon` action also demonstrate the description of request attribu
 A coupon contains information about a percent-off or amount-off discount you might want to apply to a customer.
 
 + Parameters
-    + id (string) 
+    + id (string)
 
         The ID of the desired coupon.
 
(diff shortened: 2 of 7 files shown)


API Blueprint Specification.md
```diff
@@ -766,30 +766,9 @@ Defined by the `Parameters` keyword written in a Markdown list item:
 #### Description
 Discussion of URI parameters in the _scope of the parent section_.
 
-This section **must** be composed of nested list items only. This section **must not** contain any other elements. One list item per URI parameter. The nested list items subsections inherit from the [Named section](#def-named-section) and are subject to additional formatting as follows:
+This section **must** be composed of nested list items only. This section **must not** contain any other elements. Each of the list item describes a single URI parameter using the **[Markdown Syntax for Object Notation][MSON] (MSON)**.
```
kylef: Should this read: "Each of the list items" instead of "Each of the list item".

examples/09. Advanced Attributes.md
```diff
@@ -47,7 +47,7 @@ Retrieves the coupon with the given ID.
 Returns a list of your coupons.
 
 + Parameters
-    + limit = `10` (optional, number)
+    + limit: 10 (optional, number)
```
kylef: Most of the other examples you've removed `optional` as it's optional by default. Is there any reason you didn't for this particular case?

API Blueprint Specification.md
```diff
@@ -766,30 +766,9 @@ Defined by the `Parameters` keyword written in a Markdown list item:
...
-
-* `<parameter name>` is the parameter name as written in [Resource Section](#ResourceSection)'s URI (e.g. "id").
-* `<description>` is any **optional** Markdown-formatted description of the parameter.
-* `<additional description>` is any additional **optional** Markdown-formatted [description](#SectionDescription) of the parameter.
-* `<default value>` is an **optional** default value of the parameter – a value that is used when no value is explicitly set (optional parameters only).
-* `<example value>` is an **optional** example value of the parameter (e.g. `1234`).
-* `<type>` is the **optional** parameter type as expected by the API (e.g. "number").
-* `Values` is the **optional** enumeration of possible values
-*  and `<enumeration element n>` represents an element of enumeration type.
-* `required` is the **optional** specifier of a required parameter (this is the **default**)
```
kylef: We're changing from making `required` the default, to making `optional` which I think may be a bad idea. This would be a breaking change and most likely catch people out.

zdne:
> For @zdne to review & merge.

I think it is OK if @kylef or even @klokane reviews this and merge it (I think it is not really needed to write these addresses  )

API Blueprint Specification.md
zdne: :+1:

API Blueprint Specification.md
zdne: Agree. I do not think we can realign this. The only thing I can think of is to do not make the requirement flag optional – it should be always specified otherwise parser will warn. Ideas?

API Blueprint Specification.md
```diff
@@ -801,32 +780,35 @@ Where:
...
 ```
 
 ```
 + Parameters
-    + id = `20` (optional, number, `1001`) ... Id of a post.
+    + id: `1001` (number) - Id of a post.
+        + Default: `20`
 ```
 
+You can give an enumeration of possible values using [MSON][] as shown below:
```
zdne: Please do not refer to MSON here. Explain it without referencing to MSON syntax.

API Blueprint Specification.md
zdne: Also I am not convinced that we should refer to MSON here. Align the syntax yes. But I would not claim it is MSON syntax as there are some gaps such as the required vs. optional or the fact that there are no nested types.

Also explaining URI syntax here without references to MSON would set a great building stone for introducing MSON later...

zdne:
@pksunkara reviewed. Besides @kylef comments I would say we can't really claim the URI parameter syntax IS MSON syntax, because, well, it isn't. Please take the effort and explain the syntax in words without referring to MSON.

examples/09. Advanced Attributes.md
pksunkara: This was a mistake.

pksunkara:
> I would say we can't really claim the URI parameter syntax IS MSON syntax, because, well, it isn't.

I strongly disagree with that. In my opinion, we should want to move URI parameter syntax to MSON syntax because of three things:

1. Few APIs allow arrays to be sent in a query parameter. (Ex: CouchDB API). So, we do want to allow users the option for complex parameter values.
2. We might want to use the URI parameters to improve the Request and/or Response payloads in the future (along with using them in the mock server)
3. Even though parameters won't contain more than 1 or 2 nested members, I don't want to add a separate description for them and think it is better if we just point the users to MSON. I am saying this because I think it will be easier for the users this way rather than to read different things for parameters and attributes.

API Blueprint Specification.md
pksunkara: > The only thing I can think of is to do not make the requirement flag optional – it should be always specified otherwise parser will warn.

Are you sure you are talking about the old syntax? I think what @kylef is trying to say is:

```
+ id (string)
```

In the old syntax, the parameter `id` is `required`, while in the new syntax, the parameter `id` is `optional`.

Please note that if the user adds a bit more to it like for example:

```
+ id = `1` (string)
```

or 

```
+ id: `1` (string)
```

We can easily understand which syntax he is using and can understand whether he is saying `optional` or `required`.

I think the user will not be confused when writing the blueprint. Its when the blueprint is parsed and the snowcrash is unable to decide on the old or the new syntax.

examples/09. Advanced Attributes.md
pksunkara: But this should not be a problem because even though `optional` is default, the user can still say it. I am actually thinking of leaving this as it is. What do you say?

jrep:
@pksunkara : I'm OK with *adding* fully-MSON parameter descriptions, but if your "move" means "forbid the old way" I would strongly disagree. We are using Blueprint to document legacy, often highly un-RESTful APIs (as well as documenting and developing more RESTful ones). If the Blueprint language should lose the power to do that, we couldn't continue using it.

But I agree with @zdne in that the MSON syntax for parameters must be in place first, before it can be used by reference to explain things.

pksunkara:
@jrep The old syntax will still be supported even after the move. So, you should not have a worry.

> But I agree with @zdne in that the MSON syntax for parameters must be in place first, before it can be used by reference to explain things.

I am sorry but I didn't get what you mean by this.

jrep:
About "move": OK, I'm satisfied about that.

About "agree with @zdne": 

1. You added "You can give an enumeration of possible values using [MSON][] as shown below".
1. Z commented "we can't really claim the URI parameter syntax IS MSON syntax, because, well, it isn't".
1. You responded "I strongly disagree with that", quoting Z's comment as I have done here.

I take Z's comment to mean "at the moment, URI parameter syntax is not MSON. Therefore, at the moment, we shouldn't put in a link that claims that it is MSON." Z's comment doesn't explicitly say whether the fix should be to remove the link (which would, of course, be easy), or to fix up URI parameter syntax to be MSON (as you have advocated, a more difficult but possibly higher goal). I took your "strong disagreement" to mean "it's OK to put the MSON link here, even though URI parameter syntax is not MSON." 

Perhaps you meant something lesser, like "this particular detail is already MSON-compatible, so the link is OK in this case." I would disagree with that, for fear of creating the false impression that URI parameter syntax is *entirely* MSON.

Or, perhaps you meant "I agree that URI parameter syntax is presently not MSON, but I believe we should respond by fixing URI parameter syntax actually to *be* MSON, at which point this link would be OK." Now that you've reassured me about "move" and backwards compatibility, I'm OK with this philosophical position, but I continue to hold (in agreement, I believe, with Z) that the upgrades to URI parameter syntax should come first, and only when that is complete does this link become OK.

And as background, I'm a little dubious that this upgrade of URI parameter syntax to full MSON-ness is actually practical, so I'd rather not see the link sneak in on mere speculation that MSON-ization will eventually happen.

pksunkara:
Okay, got it. Thanks for the exaplanation :smile: 

Firstly, the changes in this PR will get published only after they are implemented in the parser. So

> "I agree that URI parameter syntax is presently not MSON, but I believe we should respond by fixing URI parameter syntax actually to be MSON, at which point this link would be OK."

This is what I am trying to do.

I think what @zdne was trying to say is that, "Parameters should not be MSON, just the syntax should resemble a bit of MSON". That statement is what I strongly disagree with. I think it should be completely MSON not just resemblance.

> I'm a little dubious that this upgrade of URI parameter syntax to full MSON-ness is actually practical, so I'd rather not see the link sneak in on mere speculation that MSON-ization will eventually happen.

This should not be an issue due to how MSON was implemented in the parser. If we do decide on going the MSON route, there will be no half-measures when implementing the change in the parser.

examples/09. Advanced Attributes.md
zdne: What do you mean by "as is"? Please elaborate.

zdne:
@jrep thanks for great feedback! I think we are aligned here. Although I have a question: Do you think we should keep the deprecated syntax in the specification (or anywhere else for that matter)? The motivation behind this question is to prevent confusion for readers of older blueprints.

zdne:
@pksunkara 

> I strongly disagree with that. In my opinion, we should want to move URI parameter syntax to MSON syntax (not simply align it but complete move it) 

Let's do it step by step. First align. Then move it if it will make sense. 

> Few APIs allow arrays to be sent in a query parameter. (Ex: CouchDB API). So, we do want to allow users the option for complex parameter values.

Aligning the syntax first won't prevent supporting this in a future

> We might want to use the URI parameters to improve the Request and/or Response payloads in the future (along with using them in the mock server)

Again, the alignment won't prohibit us from doing this in future. 

> Even though parameters won't contain more than 1 or 2 nested levels, I don't want to add a separate description for them

Let's not be lazy. Aligning this is actually **much easier and safer** then referencing MSON. There is way too many exception in MSON that you would have to take care (read document) them than defining the primitives (consider On Of, Include, Inheritance, Nested members).

zdne:
Two main points here: 

1. URI Parameters syntax will remain backward compatible
2. URI Parameter syntax won't be, for the time being, full-blown MSON

API Blueprint Specification.md
zdne: Unfortunately I do not think we can now change this behavior. The URI params have to stay required by default.

In fact, we should distinguish between paths segments variables and query parameters. The first should be required, the later should be optional by default. But this is a larger project, perhaps for future.

pksunkara:
Updated.

API Blueprint Specification.md
```diff
@@ -766,17 +766,19 @@ Defined by the `Parameters` keyword written in a Markdown list item:
 #### Description
 Discussion of URI parameters in the _scope of the parent section_.
 
-This section **must** be composed of nested list items only. This section **must not** contain any other elements. One list item per URI parameter. The nested list items subsections inherit from the [Named section](#def-named-section) and are subject to additional formatting as follows:
+This section **must** be composed of nested list items only. This section **must not** contain any other elements. Each of the list items describes a single URI parameter. The nested list items subsections inherit from the [Named section](#def-named-section) and are subject to additional formatting as follows:
```
kylef: I think this could be a little simpler:

```diff
-Each of the list items describes a single URI parameter.
+Each list item describes a single URI parameter.
```

API Blueprint Specification.md
```diff
@@ -785,9 +787,10 @@ Where:
 * `<additional description>` is any additional **optional** Markdown-formatted [description](#SectionDescription) of the parameter.
 * `<default value>` is an **optional** default value of the parameter – a value that is used when no value is explicitly set (optional parameters only).
 * `<example value>` is an **optional** example value of the parameter (e.g. `1234`).
-* `<type>` is the **optional** parameter type as expected by the API (e.g. "number").
-* `Values` is the **optional** enumeration of possible values
-*  and `<enumeration element n>` represents an element of enumeration type.
+* `<type>` is the **optional** parameter type as expected by the API (e.g. "number", "string", "boolean"). "string" is the **default**.
+* `Members` is the **optional** enumeration of possible values. `<type>` should be surrounded by `enum[]` if this is present.
```
zdne: > `<type>` should be surrounded by `enum[]` if this is present.

This is probably not the best way to say what you want to say. Please feel free to elaborate more thoughtfully.

"

API Blueprint Specification.md
```diff
@@ -785,9 +787,10 @@ Where:
...
+* `<enumeration value n>` represents an element of enumeration type.
+* `<enumeration description n>` represents the description of the corresponding enumeration element.
```
zdne: You are missing "optional" here.

API Blueprint Specification.md
```diff
@@ -801,34 +804,35 @@ Where:
...
 + Parameters
-    + id (string)
+    + id (enum[string])
 
         Id of a Post
 
-        + Values
-            + `A`
+        + Members
+            + `A` - This is alphabet A
```
zdne: alphabet :arrow_right: character?

zdne:
Looks good to me. Main comment is on explaining `members` section with regard to the type value of an `enum[type]`. Please describe it more thoroughly

pksunkara:
Updated.

PR #159, opened 2015-02-24
Inline action syntax

pksunkara:
For @zdne to review & merge.

I was unable to decide about adding an example API showing the inline action in the `examples` folder. Please correct me if I am wrong, but the `examples` folder is for the major things, right?

### DIFF
--- a/API Blueprint Specification.md
+++ b/API Blueprint Specification.md
@@ -687,7 +687,7 @@ This section **should** include at least one nested [Action section](#def-action
 
 - [`0-1` URI parameters section](#def-uriparameters-section)
 
-    URI parameters defined in the scope of a Resource section apply to _any and all_ nested Action sections.
+    URI parameters defined in the scope of a Resource section apply to _any and all_ nested Action sections except when an [URI template][uritemplate] has been defined for the Action.
 
 - [`0-1` Attributes section][]
 
--- a/API Blueprint Specification.md
+++ b/API Blueprint Specification.md
@@ -668,13 +668,13 @@ Defined by an [URI template][uritemplate]:
 
 **-- or --**
 
-Defined by a resource [name (identifier)](#def-identifier) followed by an URI template enclosed in square brackets `[]`.
+Defined by a resource [name (identifier)](#def-identifier) followed by an [URI template][uritemplate] enclosed in square brackets `[]`.
 
     # <identifier> [<URI template>]
 
 **-- or --**
 
-Defined by an [HTTP request method][httpmethods] followed by URI template:
+Defined by an [HTTP request method][httpmethods] followed by [URI template][uritemplate]:
 
     # <HTTP request method> <URI template>
 
@@ -848,7 +848,13 @@ Defined by an [HTTP request method][httpmethods]:
 
 Defined by an action [name (identifier)](#def-identifier) followed by an [HTTP request method][httpmethods] enclosed in square brackets `[]`.
 
-    # <identifier> [<HTTP request method>]
+    ## <identifier> [<HTTP request method>]
+
+**-- or --**
+
+Defined by an action [name (identifier)](#def-identifier) followed by an [HTTP request method][httpmethods] and [URI template][uritemplate] enclosed in square brackets `[]`.
+
+    ## <identifier> [<HTTP request method> <URI template>]
 
 #### Description
 Definition of at least one complete HTTP transaction as performed with the parent resource section. An action section **may** consists of multiple HTTP transaction examples for the given HTTP request method.
@@ -891,6 +897,13 @@ Multiple Request and Response nested sections within one transaction example **s
     + Response 201
 
             ...
+
+    ### Delete a Post [DELETE /posts/{id}]
+
+    + Parameters
+        + id (string) ... Id of the post
+
+    + Response 204
 
 #### Example Multiple Transaction Examples
 
--- a/examples/12. Advanced Action.md
+++ b/examples/12. Advanced Action.md
@@ -1,0 +1,66 @@
+FORMAT: 1A
+
+# Advanced Action API
+An action in a resource in the API Blueprint can be considered as a state transition. This API example demonstrates that a resource can contain a state transition to another resource. This allows us to show that the resource `Task` can afford an action that is not limited to the resource `Task`.
+
+## API Blueprint
++ [Previous: Resource Model](11.%20Resource%20Model.md)
++ [This: Raw API Blueprint](https://raw.github.com/apiaryio/api-blueprint/master/examples/11.%20Advanced%20Action.md)
+
+# Task [/tasks/{id}]
+
++ Parameters
+    + id (string)
+
+## Retrieve Task [GET]
+
++ Response 200 (application/json)
+
+        {
+            "id": 123,
+            "name": "Go to gym",
+            "done": false,
+            "type": "task"
+        }
+
+## Update Task [PATCH]
+This is a state transition within the same resource
+
++ Request (application/json)
+
+        {
+            "name": "Exercise in gym"
+        }
+
++ Response 200 (application/json)
+
+        {
+            "id": 123,
+            "name": "Exercise in gym",
+            "done": false,
+            "type": "task"
+        }
+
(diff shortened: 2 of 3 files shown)


zdne:
"For @zdne to review & merge", again @kylef is fully capable of doing this as well :)

zdne:
> Please correct me if I am wrong, but the examples folder is for the major things, right?

I would say it is for any and all examples that are worthy. If they are part of the "course" they should be numbered where the lower number the more fundamental topic of the example (so maybe this is one with a fairly high number)

API Blueprint Specification.md
```diff
@@ -848,7 +848,15 @@ Defined by an [HTTP request method][httpmethods]:
 
 Defined by an action [name (identifier)](#def-identifier) followed by an [HTTP request method][httpmethods] enclosed in square brackets `[]`.
 
-    # <identifier> [<HTTP request method>]
+    ## <identifier> [<HTTP request method>]
+
+**-- or --**
+
+Defined by an action [name (identifier)](#def-identifier) followed by an [HTTP request method][httpmethods] and [URI template][uritemplate] enclosed in square brackets `[]`. This is also known by the name *Inline Action*.
```
zdne: I would remove "This is also known by the name *Inline Action*."

API Blueprint Specification.md
```diff
@@ -848,7 +848,15 @@ Defined by an [HTTP request method][httpmethods]:
...
+
+    # <identifier> [<HTTP request method> <URI template>]
+
+> **NOTE:** The inline action is still considered a part of the resource it is nested under even though it's URI template is different to that of the resource.
```
zdne: This note does not make much sense. Any and all actions – link transitions – defined under a resource are for that particular resource, regardless of where the transition goes. This is like saying "all the links on this HTML page are part of the page even though the link points elsewhere"

zdne:
Blockers are my comments above and missing example both in specification and in the examples folder. Also it would be good to explain action a resource affordance represented by link transition. Distinguishing state transition within the same resource `## <identifier> [<HTTP request method>]` and state transition that to another resource `# <identifier> [<HTTP request method> <URI template>]`. 

Feel free to spend some time studying the state transition topic as it is fundamental principle of REST architecture.

pksunkara:
Updated.

examples/12. Inline Action.md
```diff
@@ -0,0 +1,66 @@
+FORMAT: 1A
+
+# Inline Action API
```
zdne: So while we use this name "internally" I do not think we should communicate is as a (new) concept. There is no such a thing as *inlined* transitions or affordances. I would simply name this example as "Advanced Action"

examples/12. Inline Action.md
```diff
@@ -0,0 +1,66 @@
...
+This API example demonstrates that a resource can contain an in-place definition of another resource. This allows us to show that the resource `Task` can afford an action that is not limited to the resource `Task`.
```
zdne: > resource can contain an in-place definition of another resource

This is actually not correct statement. Again, this should be talking about arbitrary state transitions (precisely – state transitions that are not limited to the parent resource as its target)

zdne:
Solid kick-off. Although I think in needs a little bit of touch – Since we are introducing the term "state transition" here for the first time, if I am correct, I think it would be good to mention somewhere in the lead that Action as such is a state transition..

pksunkara:
Updated.

examples/12. Advanced Action.md
```diff
@@ -0,0 +1,66 @@
+FORMAT: 1A
+
+# Advanced Action API
+An action in a resource in the API Blueprint can be considered as a state transition. This API example demonstrates that a resource can contain a state transition to another resource. This allows us to show that the resource `Task` can afford an action that is not limited to the resource `Task`.
```
zdne: > An action in a resource in the API Blueprint can be considered as a state transition

A resource action is – in fact – a state transition. 

> This API example demonstrates that a resource can contain a state transition to another resource.

This API example demonstrates an action – state transition – to another resource. 

>   This allows us to show that the resource `Task` can afford an action that is not limited to the resource `Task`.

I would remove this sentence completely

examples/12. Advanced Action.md
```diff
@@ -0,0 +1,66 @@
...
++ Response 200 (application/json)
+
+        {
+            "id": 123,
+            "name": "Exercise in gym",
+            "done": false,
+            "type": "task"
+        }
+
+## List All Tasks [GET /tasks{?status,priority}]
```
zdne: While this is a valid example, I consider it  not the best show case. Usually an item of collection should not be "affording" you with viewing every other item – in my opinion it should be the other way around. A collection should afford you with the action to view, create, delete a single item.

zdne: Checkout the diagram at https://github.com/resource-blueprint/resource-blueprint#gist-resource I would suggest to change this example either to `Tasks` and "view a single task", "delete a single task" etc.

examples/12. Advanced Action.md
```diff
@@ -0,0 +1,66 @@
...
+
+        {
+            "id": 123,
+            "name": "Go to gym",
+            "done": false,
+            "type": "task"
+        }
+
+## Update Task [PATCH]
+This is a state transition within the same resource
```
zdne: Note this does not necessarily change the state. There is a subtle difference in between changing the values and state of a resource.

fosrias: IMO, changing data does change state. There are infinite resource states that you can access using the `edit` transition. Those states all have the same affordances, but because of the resource has changed, it has a different state. From the state-machine standpoint, there are only so many combinations of a affordances, but depending what those affordances are, there are potentially infinite 'states' in any one of them (e.g. they allow you to edit). Make sense?

zdne: Well. As you have said from the FSM there is only a finite (ha) number of states. This is what I have meant by some actions – transitions – are not changing the state (for the most basic example look at the "self" relation type).

zdne: So all in all depends on what you consider to be a state. From FSM standpoint not every action changes the state and I believe this is the state we are talking about when we are talking about transition changing the state (of FSM)

fosrias: Yeah. I have wrestled with how to articulate this. However, if you think
about it, everytime you exercise a state transition you cycle to a new
"state". The fact that on the diagram it is the "same" state category can
obscure that it is a new state. Even calling `self` is a new state on a
resource since the data may have changed since you originally received it.
I say there are infinite data states here, but only so many categories of
states where the possible affordances change (as driven by some change in
the data state). Anyhow, that is how I think about it.

kylef:
In the specification's resource section (10), there is the following statement:

> URI parameters discussed in the scope of a Resource section apply to any and all nested Action sections.

I think this needs to be reworded as I don't believe it applies anymore anymore, given the following example:

```markdown
## User [/users/{username}]

+ Parameters
    + username...

### Create a User [POST /users]
```

examples/12. Advanced Action.md
zdne: > categories of states

Oh my

zdne: Either way I think we are in an agreement here. There are two "states" definitions: 

1. a state is one particular combination of data values and available affordances
2. a state is one particular combination of available affordances regardless of data values

So I believe originally I was talking about No2 while you made your point on No1. It would is good to figure this one out!

zdne:
> I think this needs to be reworded as I don't believe it applies anymore anymore

@kylef good catch! 

@pksunkara please take this into account

pksunkara:
Updated.

examples/12. Advanced Action.md
```diff
@@ -0,0 +1,55 @@
+FORMAT: 1A
+
+# Advanced Action API
+A resource action is – in fact – a state transition. This API example demonstrates that an action - state transition - to another resource.
```
zdne: The last sentence does not make sense. Please re-read it loud & then improve it.

zdne:
Please fix the example.

pksunkara:
Oops. Forgot to remove the word `that`. Fixed now.

PR #162, opened 2015-02-26
Hypermedia Variant of the Polls Example

kylef:
This pull request adds a more advanced example blueprint showing both Siren and HAL support in the Polls API along with Resource models.

### DIFF
--- a/examples/Polls Hypermedia API.md
+++ b/examples/Polls Hypermedia API.md
@@ -3,7 +3,7 @@ HOST: http://polls.apiblueprint.org/
 
 # Polls
 
-Polls is a simple web service that allows consumers to view polls and vote in them. You can view this documentation over at [Apiary](http://docs.pollshypermedia.apiary.io/).
+Polls is a simple API allowing consumers to view polls and vote in them. You can view this documentation over at [Apiary](http://docs.pollshypermedia.apiary.io/).
 
 # Polls API Root [/]
 
@@ -30,12 +30,12 @@ This resource does not have any attributes. Instead it offers the initial API af
             }
         }
 
-## Questions collection [/questions{?page}]
+## Questions Collection [/questions{?page}]
 
 + Parameters
     + page (optional, number, `1`) ... The page of questions to return
 
-### List all questions [GET]
+### List All Questions [GET]
 
 + Response 200 (application/vnd.siren+json)
 
@@ -217,9 +217,9 @@ This resource does not have any attributes. Instead it offers the initial API af
             }
         }
 
-### Create a new question [POST]
+### Create a New Question [POST]
 
-You can create your own question using this action. It takes a JSON dictionary containing a question and a collection of answers in the form of choices.
+You may create your own question using this action. It takes a JSON dictionary containing a question and a collection of answers in the form of choices.
 
 - question (string) - The question
 - choices (array[string]) - A collection of choices.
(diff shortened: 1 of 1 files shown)


examples/Polls Hypermedia API.md
```diff
@@ -0,0 +1,408 @@
...
+
+## Retrieve the Entry Point [GET]
+
++ Response 200 (application/vnd.siren+json)
+
+        {
+            "links": [
+                {
+                    "rel": [ "questions" ],
+                    "href": "\/questions"
```
zdne: Is the `\` needed?

examples/Polls Hypermedia API.md
```diff
@@ -0,0 +1,408 @@
...
+                }
+            ]
+        }
+
++ Response 200 (application/hal+json)
```
zdne: For this to work you want to add 

```
+ Request 
    + Headers

            Accept: application/hal+json
```

zdne: And Siren's version as well (above the Siren response)

examples/Polls Hypermedia API.md
```diff
@@ -0,0 +1,408 @@
...
+
+- question
+- published_at - An ISO8601 date when the question was published.
+- url
+- choices - An array of Choice objects.
+
++ Parameters
+    + question_id (required, number, `1`) ... ID of the Question in form of an integer
+
++ Model (application/vnd.siren+json)
```
zdne: As I want to deprecate and remove the Model concept entirely (to be replaced with Attributes) I would not use it in this tutorial..

examples/Polls Hypermedia API.md
kylef: Guess not, was using a serialiser that escapes them. Will strip it out.

examples/Polls Hypermedia API.md
kylef: Is it required to explicitly put this? I though it would be a given that [content negotiation](http://www.w3.org/Protocols/rfc2616/rfc2616-sec12.html) as per HTTP 1.1 specification would work to get the expected result.

(I am happy to add it if you see this is needed).

zdne: Hmm thinking about this maybe you are right. Let's keep it simple.

zdne:
@pksunkara @kylef guys since you have merged #164 this can no longer got to the master...

zdne:
Also if this should conform to the latest development (since you have added relations) it should use the new Parameters syntax. 

At the current state this is a mess.

zdne:
Note: as a consequence of having #164 in this, this prohibits the whole new Tutorial story to be released without attributes being released. This is not good. Please untangle these two for now.

kylef:
Good point, just reverted that merge.

examples/Polls Hypermedia API.md
```diff
@@ -0,0 +1,613 @@
...
+                    }
+                ]
+            }
+        }
+
+### Create a New Question [POST]
+
+You may create your own question using this action. It takes a JSON object containing a question and a collection of answers in the form of choices.
+
+- question (string) - The question
```
zdne: I still have concerns mixing `-`s with `+`s (as to prevent confusion), but I will leave this up to your judgement @kylef

zdne:
@kylef Merging, if you would like to unify `-`s and `+`s I think it would be good...

PR #154, opened 2015-02-19
Simpler API Blueprint Tutorial

kylef:
These changes make the new tutorial shorter, and simpler while requiring less prerequisites. The tutorial has no expectations of the user knowing Markdown.

Would appreciate any feedback on this one to ensure we get this right.

PR #163, opened 2015-03-02
Address feedback from polls tutorial in the example

kylef:
This pull request includes changes for the feedback from #154.

- Use may not can.
- Use titlecase

### DIFF
--- a/examples/Polls API.md
+++ b/examples/Polls API.md
@@ -124,7 +124,7 @@ This action allows you to vote on a question's choice.
 
 ### Create a New Question [POST]
 
-You may create your own question using this action. It takes a JSON dictionary containing a question and a collection of answers in the form of choices.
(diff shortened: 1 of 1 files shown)


examples/Polls API.md
```diff
@@ -80,12 +80,12 @@ This action allows you to vote on a question's choice.
...
             Location: /questions/1
 
-## Questions collection [/questions{?page}]
+## Questions Collection [/questions{?page}]
 
 + Parameters
     + page (optional, number, `1`) ... The page of questions to return
 
-### List all questions [GET]
+### List All Questions [GET]
```
pksunkara: Are we sure about this? In apiary docs, this is an action name and this would still be title case.

zdne: @pksunkara I do not see any problem here. Please elaborate if this is the case.

PR #174, opened 2015-03-23
[Examples] Refer to resources group as plural

kylef:
This is from @zdne's feedback on https://github.com/apiaryio/apiary/pull/2203#discussion_r26917091


zdne:
:+1:

PR #178, opened 2015-04-07
Update API Blueprint Specification.md

adammbalogh:
Attributes section fix.

I think the body should be indented by 12 space (or 3 tabs) in this case.


kylef:
Thanks @adammbalogh

PR #184, opened 2015-04-27
Update tutorial with new-style params

danielgtaylor:
As described here:

https://github.com/apiaryio/api-blueprint/blob/master/API%20Blueprint%20Specification.md#13-uri-parameters-section


zdne:
:+1:

Issue #196, opened 2015-05-19
Named resource with method and uri


LinusU:
I'm trying to write a quick start guide in addition to the api specification that will help new users to quickly get started. I'm running into problems when trying to name the steps in the tutorial.

This is the blueprint:

```apib
# Group Quick start

## Create user [POST /users]

The first thing we need to do is to add a user that we can later log in as. Do
this by sending a POST request with the relevant details.

+ Request
  + Headers
    Content-Type: application/json
  + Body
    { "username": "Linus", "password": "test" }

+ Response 201
  + Headers
    Location: /user/1337

## Log in as the user [POST /session]

We can now log in as the user we created. This is accomplished by creating a new
session with the relevant login details.

+ Request
  + Headers
    Content-Type: application/json
  + Body
    { "username": "Linus", "password": "test" }

+ Response 201
  + Headers
    X-Token: djksaldjaklsj894ut93
    Location: /session/8493248
```

The problem is `## Create user [POST /users]` and `## Log in as the user [POST /session]`.

For some reson they gets parsed as a text block instead of a resource with one action. If I remove the name I gave them, and only add the method + uri, it works as expected. (i.e. `## POST /users`)

Shouldn't this work?

zdne:
Unfortunately no :( Currently only actions under a resource can be named – the idea was to encourage resource oriented architectures instead of just writing endpoints. 

However what you are suggesting is making a sense (and I actually really like it). We are looking at "scenario" support as a part of testing – https://github.com/apiaryio/api-blueprint/issues/21 And this is somewhat related. 

What do you think? I would suggest to have a look at the scenarios in #21 and refine a syntax proposal from this. Does it make sense. Or is it really just "named" endpoints that would help here and scenarios are overkill?

LinusU:
Hmm, as far as I understood the scope of API Blueprint it's not only meant for REST APIs. For example the website says "API Documentation with powerful tooling" and dosen't mention REST at all.

If this is the case I think it's very bad to "encourage resource oriented architectures" as some APIs just aren't fit for that architecture. I think that API Blueprint should treat every api design as equals and not try and nudge people against one of them.

Scenarios is very cool, but it seems very testing oriented. While this is great for my needs, I'm not sure that API Blueprint should bend that much towards testing. I think that the approach that I took in [LinusU/mocha-api-blueprint-example](https://github.com/LinusU/mocha-api-blueprint-example) is more of a documentation first approach.

[...]

PR #198, opened 2015-05-26
spec: allow named endpoints

LinusU:
See discussion in #196

### DIFF
--- a/API Blueprint Specification.md
+++ b/API Blueprint Specification.md
@@ -534,7 +534,13 @@ Defined by an [HTTP request method][httpmethods] followed by [URI template][urit
 
     # <HTTP request method> <URI template>
 
-> **NOTE:** In this case the rest of this section represents the [Action section](#def-action-section) including its description and nested sections and **follows the rules of Action section instead**.
+**-- or --**
+
+Defined by a resource [name (identifier)](#def-identifier) followed by an [HTTP request method][httpmethods] and an [URI template][uritemplate] enclosed in square brackets `[]`:
+
+    # <identifier> [<HTTP request method> <URI template>]
+
+> **NOTE:** In the two latter cases the rest of this section represents the [Action section](#def-action-section) including its description and nested sections and **follows the rules of Action section instead**.
 
 #### Description
 An API [resource](http://www.w3.org/TR/di-gloss/#def-resource) as specified by its *URI* or a set of resources (a resource template) matching its *URI template*.
--- a/API Blueprint Specification.md
+++ b/API Blueprint Specification.md
@@ -1,12 +1,12 @@
 ---
 
 Author: z@apiary.io
-Version: 1A8
+Version: 1A9
 
 ---
 
 # API Blueprint
-#### Format 1A revision 8
+#### Format 1A revision 9
 
 ## [I. API Blueprint Language](#def-api-blueprint-language)
 + [Introduction](#def-introduction)
--- a/API Blueprint Specification.md
+++ b/API Blueprint Specification.md
@@ -540,7 +540,7 @@ Defined by a resource [name (identifier)](#def-identifier) followed by an [HTTP 
 
     # <identifier> [<HTTP request method> <URI template>]
 
-> **NOTE:** In the two latter cases the rest of this section represents the [Action section](#def-action-section) including its description and nested sections and **follows the rules of Action section instead**.
+> **NOTE:** In the latter two cases the rest of this section represents the [Action section](#def-action-section) including its description and nested sections and **follows the rules of Action section instead**.
 
 #### Description
 An API [resource](http://www.w3.org/TR/di-gloss/#def-resource) as specified by its *URI* or a set of resources (a resource template) matching its *URI template*.
--- a/examples/13. Named Endpoints.md
+++ b/examples/13. Named Endpoints.md
@@ -1,0 +1,42 @@
+FORMAT: 1A
+
+# Named Endpoints API
+This API example demonstrates how to define a standalone endpoint with an identifier.
+
+## API Blueprint
++ [Previous: Advanced Action](12.%20Advanced%20Action.md)
++ [This: Raw API Blueprint](https://raw.github.com/apiaryio/api-blueprint/master/examples/13.%20Named%20Endpoints.md)
+
+# Group Quick start
+
+## Create message [POST /messages]
+
+Start out by creating a message for the world to see.
+
++ Request (application/json)
+
+        { "message": "Hello World!" }
+
++ Response 201
+
+    + Headers
+
+            Location: /messages/1337
+
+## Create a new task [POST /tasks]
+
+Now create a task that you need to do at a later date.
+
++ Request (application/json)
+
+        {
+            "name": "Exercise in gym",
+            "done": false,
+            "type": "task"
+        }
+
++ Response 201
+
+    + Headers
+
+            Location: /tasks/1992
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ API Blueprint, its parser, and most of its tools are completely open sourced so 
 
 ## Format support
 
-* Current format: [1A8](https://github.com/apiaryio/api-blueprint/releases/tag/format-1A8)
+* Current format: [1A9](https://github.com/apiaryio/api-blueprint/releases/tag/format-1A9)
 * Format supported by Apiary: [1A8](https://github.com/apiaryio/api-blueprint/releases/tag/format-1A8)
 
 ## Getting started with API Blueprint
--- a/examples/12. Advanced Action.md
+++ b/examples/12. Advanced Action.md
@@ -6,6 +6,7 @@ A resource action is – in fact – a state transition. This API example demons
 ## API Blueprint
 + [Previous: Resource Model](11.%20Resource%20Model.md)
 + [This: Raw API Blueprint](https://raw.github.com/apiaryio/api-blueprint/master/examples/12.%20Advanced%20Action.md)
++ [Next: Named Endpoints](13.%20Named%20Endpoints.md)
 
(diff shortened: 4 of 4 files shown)


zdne:
When I think about it more conceptually – the endpoint as such is actually an action section in API Blueprint parlance (albeit it implies a new resource). Therefore it should be defined under Action and not resource as such. 

This, however, has some implications – suddenly we can have "action" section at the top of a blueprint so the "Blueprint document structure" needs to be updated accordingly. 

Internally, in the parser, this should be still picked up by Resource parser as it essentially creates a new resource (in AST) but this is opaque to user. 

Furthermore parser should not warn when an action with the same URI template (but different method) is described, instead, it should add it under the respective resource. 

Overall as simply as it looks it is actually quite complex problem :)

LinusU:
Hmm, I guess that in that case we'd want to change the behaviour of `# GET /e1` as well, no?

I'm not sure if I'm happy with letting content "jump" to different places thought, it would make the internals quite a bit more complicated?

It would also make it complicated to do "guides" with blueprint since the content might not necessarily come in the order that you specify it in the markdown.

e.g.
```markdown
# Create a user [POST /users]

...

# Log in as the user [POST /sessions]

...

# List all users [GET /users]

...
```

This would now list `List all users` before it listed `Log in as the user` making that call fail...

examples/13. Named Endpoints.md
```diff
@@ -0,0 +1,41 @@
...
+# Group Quick start
+
+## Create user [POST /users]
+
+The first thing we need to do is to add a user that we can later log in as. Do
+this by sending a POST request with the relevant details.
+
++ Request
+  + Headers
+    Content-Type: application/json
```
pksunkara: The content inside `Headers` and `Body` list items should be code blocks. So, indent them by 12 spaces.

examples/13. Named Endpoints.md
```diff
@@ -0,0 +1,41 @@
...
+
+# Group Quick start
+
+## Create user [POST /users]
+
+The first thing we need to do is to add a user that we can later log in as. Do
+this by sending a POST request with the relevant details.
+
++ Request
+  + Headers
```
pksunkara: The second-level list items should be indented by 4 spaces according to our API Blueprint examples style guide.

pksunkara:
A few styling issues other than looks good. But I agree with @zdne that even though this issue looks simple, it is a bit complex one. We need to do some thinking on this.

API Blueprint Specification.md
```diff
@@ -534,7 +534,13 @@ Defined by an [HTTP request method][httpmethods] followed by [URI template][urit
 
     # <HTTP request method> <URI template>
 
-> **NOTE:** In this case the rest of this section represents the [Action section](#def-action-section) including its description and nested sections and **follows the rules of Action section instead**.
+**-- or --**
+
+Defined by a resource [name (identifier)](#def-identifier) followed by an [HTTP request method][httpmethods] and an [URI template][uritemplate] enclosed in square brackets `[]`:
+
+    # <identifier> [<HTTP request method> <URI template>]
```
pksunkara: By the way, if we are doing this, please add some info about this to the section title `Header Keywords`.

zdne:
Hey @LinusU I was thinking about it a bit more and I think you are indeed right with: 

> I guess that in that case we'd want to change the behaviour of # GET /e1 as well, no?

Now doing that and also addressing 

>  the content might not necessarily come in the order that you specify it in the markdown.

is quite radical change... So lets recap on what are the options:

## 1. Add name to an endpoint

Simply allow adding a name to `# <HTTP request method> <URI template>` action and leave everything else as is. 

### Pros
- simplest way to add a name
- doesn't change current behavior

### Cons
- everything stays under one resource - you will loose the order of these "named endpoints" in the case when they are hitting the same URI

## 2. Separate endpoints from resources

Introduce a new top-level "endpoint" section – for `# <HTTP request method> <URI template>` or `# <Name> [<HTTP request method> <URI template>]`. 

### Pros
- maintain order of endpoints
- clearly distinguish between resource and endpoint architecture (or document structure)

### Cons
- need to change AST
- changing behavior of existing things
- much more complex on implementation side
- affects all the tools

---

Thoughts ?

LinusU:
@zdne great writeup.

Number 1 is what I have implemented in this pull request, the only change really is to allow a name to the resource that gets generated.

I do like number 2 quite much thought, but as you say it's much more work involved. I think that we need to clarify a bit an endpoint is. Is it a path (e.g. `/cars`) or a method + path (e.g. `GET /cars`)? Would an endpoint contain an action? Or maybe any number of action? Or is the endpoint in itself an action?

If we try and work our way from the ast, which I think is a good approach, how would the new ast structure look?

I guess that it looks something like this currently:

- `category` *root*
  - `resource`
    - `action`
  - `category`
    - `resource`
      - `action`

And it would become something like this if we added endpoints:

- `category` *root*
  - `endpoint`
  - `resource`
    - `action`
  - `category`
    - `endpoint`
    - `resource`
      - `action`

Or maybe `resource` really just should be a `category` and `action` should be an `endpoint`. The uri would than be moved and placed in every `endpoint` instead of the one `resource`.

- `category` *root*
  - `endpoint`
  - `category`
    - `endpoint`

Very simple ast at least. Every endpoint should then contain `uriTemplate` and `method`.

Maybe a good approach would be to implement number 1 now since that dosen't change anything for the existing ecosystem, and then try and figure out what we think about number 2.

I really like the ideas in number 2 thought, it makes it possible to specify APIs that aren't RESTful which I think is important.

zdne:
@LinusU OK so let's break it into two parts

1. Just adding names to top-level method + URI pairs
2. Changing the internals to reflect ordering

zdne:
> I think that we need to clarify a bit an endpoint is.

:+1: 

In the current context I am using the term "endpoint" for a method + URI pair

>  it makes it possible to specify APIs that aren't RESTful which I think is important.

:+1: 

---

Note @smizell is currently working [Refract Resource Namespace](https://github.com/refractproject/refract-spec/blob/master/namespaces/resource-namespace.md) – this eventually should be the underlaying data structure – I believe he will address these main concept there (resources and transition) using some sort of recursive pattern. 

TL;DR let's figure this one (point 2 as mentioned above) in the Resource namespace first and focus on just adding names to currently un-named method URI pairs.

Sounds good?

API Blueprint Specification.md
zdne: @pksunkara I think it is OK to leave it as it is, because it is being already covered by "Combinations of an HTTP method and URI Template (e.g. GET /resource/{id})"

examples/13. Named Endpoints.md
zdne: :+1:

examples/13. Named Endpoints.md
zdne: :+1:

examples/13. Named Endpoints.md
```diff
@@ -0,0 +1,41 @@
...
+    Content-Type: application/json
+  + Body
+    { "username": "Linus", "password": "test" }
```
zdne: Consistence: Can we please use the same / similar example data as in the previous examples?

examples/13. Named Endpoints.md
```diff
@@ -0,0 +1,41 @@
...
++ Response 201
+  + Headers
+    Location: /user/1337
+
+## Log in as the user [POST /session]
+
+We can now log in as the user we created. This is accomplished by creating a new
+session with the relevant login details.
+
++ Request
```
zdne: Did you leave Content-Type – `application/json` out on purpose ?

examples/13. Named Endpoints.md
```diff
@@ -0,0 +1,41 @@
...
+  + Headers
+    Content-Type: application/json
```
zdne: Oh never mind. But I would rather see

```
+ request (application/json)
```

examples/13. Named Endpoints.md
```diff
@@ -0,0 +1,41 @@
...
+  + Body
+    { "username": "Linus", "password": "test" }
+
++ Response 201
+  + Headers
```
zdne: Indentation

examples/13. Named Endpoints.md
zdne: doing so you won't need `+ body `

zdne:
@LinusU Please address the review comments on the example and it is good to go! Thanks! :balloon:

LinusU:
@zdne Should be good to go :+1:

API Blueprint Specification.md
```diff
@@ -534,7 +534,13 @@ Defined by an [HTTP request method][httpmethods] followed by [URI template][urit
...
+
+> **NOTE:** In the two latter cases the rest of this section represents the [Action section](#def-action-section) including its description and nested sections and **follows the rules of Action section instead**.
```
pksunkara:
```diff
- In the two latter
+ In the latter two
```

LinusU:
@pksunkara Done!

pksunkara:
Looks good. I am toggling this as "Passed Review". :+1:

pksunkara:
@zdne Since I merge apiaryio/snowcrash#332, Do we merge this? Or should we wait for releasing it on Apiary?

zdne:
@pksunkara I think we can merge this / update README (version supported in apiary) and release new version of API Blueprint...

pksunkara:
@LinusU Can you please make a new commit to this PR updating the current format to 1A9 from 1A8? Thanks.

LinusU:
I'm on it!

LinusU:
Done, I did not tag it since it seems that you have always tagged the merge commit :)

LinusU:
Cheers :beers:

Issue #11, opened 2013-08-14
Authentication


zdne:
Direct support for API authentication.
## RFC proposals
- [API Blueprint RFC 0002: Authentication Framework](https://github.com/apiaryio/api-blueprint-rfcs/blob/master/draft/0002-authentication.md)
- [API Blueprint RFC 0003: Basic Authentication Scheme](https://github.com/apiaryio/api-blueprint-rfcs/blob/master/draft/0003-authentication-basic.md)
- [PR: OAuth 2 Authentication Scheme](https://github.com/apiaryio/api-blueprint-rfcs/pull/6)

netmilk:
The [`oauthd`](https://github.com/oauth-io/oauthd) project has a very nice gallery of declarative descriptions of various OAuth authentications of more than 100+ production APIs. Have a look to files `conf.json` and `settings.json` in some of folders [here](https://github.com/oauth-io/oauthd/tree/master/providers).

I think it's a good inspiration for OAuth requirements in the API Blueprint syntax.

pksunkara:
Syntax proposed at #201

zdne:
Added RFC proposals to the OP: https://github.com/apiaryio/api-blueprint/issues/11#issue-18055746

hsluoyz:
Any update?

PR #201, opened 2015-06-01
Authentication syntax proposal

pksunkara:
This PR includes syntax proposal for Authentication support (#11) in API Blueprint. The syntax adds an `Authentication` section and an `Authentication Schemes` section to the blueprint using which the writer can define the authentication schemes supported by their API and which endpoints need which kind of authentication.

For now, an user can select authentication schemes from a list of schemes which are inherently supported by the API Blueprint. They are:

* Basic Authentication Scheme
* Oauth 2.0 Authentication Scheme
* Bearer Authentication Scheme

For more information, read the spec and the Authentication example in `examples` folder.

### DIFF
--- a/API Blueprint Specification.md
+++ b/API Blueprint Specification.md
@@ -98,6 +98,7 @@ All of the blueprint sections are optional. However, when present, a section **m
         + [`0-1` **URI Parameters** section](#def-uriparameters-section)
         + [`0-1` **Attributes** section](#def-attributes-section)
         + [`0+` **Request** sections](#def-request-section)
+            + [`0-1` **Authentication** section](#def-authentication-section)
             + [`0-1` **Headers** section](#def-headers-section)
             + [`0-1` **Attributes** section](#def-attributes-section)
             + [`0-1` **Body** section](#def-body-section)
@@ -489,7 +490,7 @@ Name and description of the API
 
 <a name="def-authentication-section"></a>
 ## Authentication section
-- **Parent sections:** none, [Action section](#def-action-section)
+- **Parent sections:** none, [Action section](#def-action-section), [Request section](#def-request-section)
 - **Nested sections:** [`1` Authentication schemes section](#def-authenticationschemes-section), [`1` Response section](#def-response-section)
 - **Markdown entity:** header, list
 - **Inherits from**: none
@@ -505,7 +506,7 @@ Defined by the `Authentication` keyword in Markdown list entity followed by opti
 
     + Authentication: <list>
 
-> **NOTE:** The second notation is only used when the parent section is an [Action section](#def-action-section).
+> **NOTE:** The second notation is only used when the parent section is either an [Action section](#def-action-section) or a [Request section](#def-request-section).
 
 #### Description
 Based on the parent section, the usage of this section differs as shown below.
@@ -532,14 +533,14 @@ The [Authentication schemes section](#def-authenticationschemes-section) defines
               "type": "error"
             }
 
-#### Action Authentication section
-Description of the authentication support for a particular action in the API.
+#### Request Authentication section
+Description of the request level authentication settings.
 
-If defined, the respecitve action of the [Action section](#def-action-section) in the API can only be accessed using a vaild authentication scheme. If a list of scheme names is provided after the `Authentication` keyword, the respective action in the API does not allow other authentication schemes.
+If defined inside a [Request section](#def-request-section) of a HTTP transaction example, it implies that the API's response will be the same as the respective response when a valid authentication is sent along with the above mentioned request. This allows another HTTP transaction example to be specified for the cases where an authentication is not needed.
 
 This section **should** not have any nested sections.
 
-##### Example
+If a list of scheme names is provided after the `Authentication` keyword, then only those particular authentication schemes will be allowed for the respective HTTP transaction.
 
 ```
 + Authentication
@@ -548,6 +549,44 @@ This section **should** not have any nested sections.
 ```
 + Authentication: basic, token
 ```
+
+##### Example
+
+    ## GET /users/bond
+
+    + Response 200 (application/json)
+
+            {
+                "username": "bond"
+            }
+
+    + Request
+        + Authentication
+
+    + Response 200 (application/json)
+
+            {
+                "username": "bond",
+                "email": "james@bo.nd",
+                "bio": "I am 007"
+            }
+
+#### Action Authentication section
+Description of the action level authentication settings.
+
+If defined, all the [Request sections](#def-requestion-section) of the respective [Action section](#def-action-section) inherits the authentication settings unless specified otherwise.
+
+This section **should** not have any nested sections.
+
+##### Example
+
+    ## Retrieve User [GET]
+
+    + Authentication
+
+    + Response 200
+
+            { ... }
 
 ---
 
@@ -585,7 +624,7 @@ Where:
 
 Description of the base authentication schemes inherently defined in API Blueprint are defined below.
 
-#### Basic Auth Scheme
+#### Basic Authentication Scheme
 This is the [Basic Authentication Scheme](http://tools.ietf.org/html/rfc2617#section-2) as specified in [RFC 2617](http://tools.ietf.org/html/rfc2617).
 
 The customizable options for this scheme are:
@@ -599,7 +638,7 @@ The customizable options for this scheme are:
         + sample_username: `bond`
         + sample_password: `Iam007`
 
-#### Oauth2 Auth Scheme
+#### Oauth2 Authentication Scheme
 This scheme uses the [Oauth2 Authorization Framework](http://tools.ietf.org/html/rfc6749) specified in [RFC 6749](http://tools.ietf.org/html/rfc6749) to get an authorization grant and uses it access the API.
 
 The customizable options for this scheme are:
@@ -609,6 +648,7 @@ The customizable options for this scheme are:
 + `token_header_keyword` represents the string used instead of "Bearer" in "Authorization" header field according to [RFC 6750](http://tools.ietf.org/html/rfc6750#section-2.1). "Bearer" is the **default**.
 + `token_body_keyword` represents the parameter name used instead of "access_token" in the body of an API request according to [RFC 6750](http://tools.ietf.org/html/rfc6750#section-2.2). "access_token" is the **default**.
 + `token_query_keyword` represents the parameter name used instead of "access_token" in the query of an API request according to [RFC 6750](http://tools.ietf.org/html/rfc6750#section-2.3). "access_token" is the **default**.
++ `sample_token` represens a sample access token value. "sample_token" is the **default**.
 
 ##### Example
 
@@ -616,12 +656,13 @@ The customizable options for this scheme are:
         + access_token_endpoint: `/token`
         + token_header_keyword: `token`
 
-#### Bearer Auth Scheme
+#### Bearer Authentication Scheme
 This scheme is popularly used when API consumers has access to their own unique access token which can be used as a bearer token as specified in [RFC 6750](https://tools.ietf.org/html/rfc6750#section-2.1).
 
 The customizable options for this scheme are:
 
 + `keyword` represents the string used instead of "Bearer" in the "Authorization" header field according to [RFC 6750](https://tools.ietf.org/html/rfc6750#section-2.1). "Bearer" is the **default**.
++ `sample_token` represents a sample bearer token value. "sample_token" is the **default**.
 
 ##### Example
 
@@ -897,7 +938,7 @@ Multiple Request and Response nested sections within one transaction example **s
 <a name="def-request-section"></a>
 ## Request section
 - **Parent sections:** [Action section](#def-action-section)
-- **Nested sections:** [Refer to payload section](#def-payload-section)
+- **Nested sections:** [Refer to payload section](#def-payload-section), [`0-1` Authentication section](#def-authentication-section)
 - **Markdown entity:** list
 - **Inherits from**: [Payload section](#def-payload-section)
 
--- a/API Blueprint Specification.md
+++ b/API Blueprint Specification.md
@@ -46,7 +46,7 @@ Version: 1A8
 + [Data Structures section](#def-data-structures)
 + [Relation section](#def-relation-section)
 + [Authentication section](#def-authentication-section)
-+ [Authentication schemes section](#def-authenticationschemes-section)
++ [Authentication Schemes section](#def-authentication-schemes-section)
 + [Authenticated section](#def-authenticated-section)
 
 ## [III. Appendix](#def-appendix)
@@ -82,7 +82,7 @@ All of the blueprint sections are optional. However, when present, a section **m
 + [`0-1` **Metadata** section](#def-metadata-section)
 + [`0-1` **API Name & overview** section](#def-api-name-section)
 + [`0-1` **Authentication** section](#def-authentication-section)
-    + [`1` **Authentication schemes** section](#def-authenticationschemes-section)
+    + [`1` **Authentication Schemes** section](#def-authentication-schemes-section)
     + [`1` **Response** section](#def-response-section)
 + [`0+` **Resource** sections](#def-resource-section)
     + [`0-1` **Authenticated** section](#def-authenticated-section)
@@ -110,6 +110,7 @@ All of the blueprint sections are optional. However, when present, a section **m
             + [`0-1` **Body** section](#def-body-section)
             + [`0-1` **Schema** section](#def-schema-section)
 + [`0+` **Resource Group** sections](#def-resourcegroup-section)
+    + [`0-1` **Authenticated** section](#def-authenticated-section)
     + [`0+` **Resource** sections](#def-resource-section) (see above)
 + [`0+` **Data Structures** section](#def-data-structures)
 
@@ -493,21 +494,21 @@ Name and description of the API
 <a name="def-authentication-section"></a>
 ## Authentication section
 - **Parent sections:** none
-- **Nested sections:** [`1` Authentication schemes section](#def-authenticationschemes-section), [`1` Response section](#def-response-section)
+- **Nested sections:** [`1` Authentication Schemes section](#def-authentication-schemes-section), [`1` Response section](#def-response-section)
 - **Markdown entity:** header
 - **Inherits from**: none
 
 #### Definition
 Defined by the `Authentication` keyword in Markdown header entity.
 
-    # Authentication
+    ## Authentication
 
 #### Description
 Authentication settings for the whole API.
 
-This section **should** include one [Authentication schemes section](#def-authenticationschemes-section) and one [Response section](#def-response-section).
+This section **should** include one [Authentication Schemes section](#def-authentication-schemes-section) and one [Response section](#def-response-section).
 
-The [Authentication schemes section](#def-authenticationschemes-section) defines all the types of authentication schemes supported by the API. The [Response section](#def-response-section) describes the error message returned in the case of failed authentication.
+The [Authentication Schemes section](#def-authentication-schemes-section) defines all the types of authentication schemes supported by the API. The [Response section](#def-response-section) describes the error message returned in the case of failed authentication.
 
 ##### Example
 
@@ -517,7 +518,7 @@ The [Authentication schemes section](#def-authenticationschemes-section) defines
     + Authentication Schemes
         + ...
 
-    + Response 403 (application/json)
+    + Response 401 (application/json)
 
             {
               "message": "Authentication required",
@@ -526,8 +527,8 @@ The [Authentication schemes section](#def-authenticationschemes-section) defines
 
 ---
 
-<a name="def-authenticationschemes-section"></a>
-## Authentication schemes section
+<a name="def-authentication-schemes-section"></a>
+## Authentication Schemes section
 - **Parent sections:** [Authentication section](#def-authentication-section)
 - **Nested sections:** none
 - **Markdown entity:** list
@@ -559,7 +560,7 @@ Where:
 + `<scheme name>` is the authentication scheme name by which this scheme is represented in [Authenticated section](#def-authenticated-section).
 + `<description>` is any **optional** Markdown-formatted description of the authentication scheme.
 + `<base scheme type>` is the base authentication scheme type (e.g. "Basic Authentication Scheme"). See below for all the available base authentication schemes.
-+ `<customizable option n>` represents a customizable option in the base authentication scheme. It is a [MSON Property Member Declaration][] without the [MSON Type Definition][] where the [MSON Property Name][] represents the name of the option and [MSON Value][] represents the value of the option.
++ `<customizable option n>` represents a customizable option in the base authentication scheme. It is a [MSON Property Member Type][] without the [MSON Type Definition][] where the [MSON Property Name][] represents the name of the option and the value of the property member represents the value of the option.
 
 List of the base authentication schemes supported by API Blueprint are described below.
 
@@ -574,32 +575,32 @@ The customizable options for this scheme are:
 ##### Example
 
     + basic (Basic Authentication Scheme)
-        + example_username: `john`
-        + example_password: `password`
+        + `example_username`: `john`
+        + `example_password`: `password`
 
 #### OAuth2 Authentication Scheme
 This scheme uses the [OAuth2 Authorization Framework](http://tools.ietf.org/html/rfc6749) specified in [RFC 6749](http://tools.ietf.org/html/rfc6749) to get an authorization grant which is then used to access the API.
 
 The customizable options for this scheme are:
 
-+ `oauth_resource_endpoint` is the oauth resource endpoint where the oauth framework resides. Value type is **string**. "/" is the **default**.
++ `resource_endpoint` is the OAuth resource endpoint where the OAuth framework resides. Value type is **string**. "/" is the **default**.
 + `scopes` represents the OAuth scopes which indicate a list of permissions. Value type is **array**. By **default** no scopes are specified.
 + `authorization_endpoint` is the authorization endpoint required for the authorization grant. Value type is **string**. "/authorize" is the **default**.
 + `access_token_endpoint` is the token endpoint used by client to obtain the access token. Value type is **string**. "/access_token" is the **default**.
-+ `revocation_endpoint` is the revocation endpoint used by client to revoke a token. Value type is **string**. "/revoke" is the **default**.
++ `revocation_endpoint` is the revocation endpoint used by client to revoke a token as defined in [RFC 7009](https://tools.ietf.org/html/rfc7009). Value type is **string**.
 + `token_header_keyword` represents the string used instead of "Bearer" in "Authorization" header field according to [RFC 6750](http://tools.ietf.org/html/rfc6750#section-2.1). Value type is **string**. "Bearer" is the **default**.
-+ `token_body_keyword` represents the parameter name used instead of "access_token" in the body of an API request according to [RFC 6750](http://tools.ietf.org/html/rfc6750#section-2.2). Value type is **string**. "access_token" is the **default**.
-+ `token_query_keyword` represents the parameter name used instead of "access_token" in the query of an API request according to [RFC 6750](http://tools.ietf.org/html/rfc6750#section-2.3). Value type is **string**. "access_token" is the **default**.
++ `token_body_keyword` represents the parameter name used instead of "access_token" in the body of an API request according to [RFC 6750](http://tools.ietf.org/html/rfc6750#section-2.2). Value type is **string**.
++ `token_query_keyword` represents the parameter name used instead of "access_token" in the query of an API request according to [RFC 6750](http://tools.ietf.org/html/rfc6750#section-2.3). Value type is **string**.
 + `example_token` represents an example access token value. Value type is **string**. "example_token" is the **default**.
 
-> **NOTE:** The value for `oauth_resource_endpoint` can include hostnames. `authorization_endpoint` and `access_token_endpoint` allow it too only if `oauth_resource_endpoint` is not specified.
+> **NOTE:** The value for `resource_endpoint` can include hostnames. `authorization_endpoint` and `access_token_endpoint` allow it too only if `resource_endpoint` is not specified.
 
 ##### Example
 
     + oauth (OAuth2 Authentication Scheme)
-        + oauth_resource_endpoint: `/oauth`
-        + access_token_endpoint: `/token`
-        + token_header_keyword: `token`
+        + `resource_endpoint`: `/oauth`
+        + `access_token_endpoint`: `/token`
+        + `token_header_keyword`: `token`
         + scopes: read, write, email
 
 #### Bearer Authentication Scheme
@@ -607,13 +608,15 @@ This scheme represents the method of sending a bearer token as specified in [RFC
 
 The customizable options for this scheme are:
 
-+ `keyword` represents the string used instead of "Bearer" in the "Authorization" header field according to [RFC 6750](https://tools.ietf.org/html/rfc6750#section-2.1). Value type is **string**. "Bearer" is the **default**.
++ `token_header_keyword` represents the string used instead of "Bearer" in "Authorization" header field according to [RFC 6750](http://tools.ietf.org/html/rfc6750#section-2.1). Value type is **string**. "Bearer" is the **default**.
++ `token_body_keyword` represents the parameter name used instead of "access_token" in the body of an API request according to [RFC 6750](http://tools.ietf.org/html/rfc6750#section-2.2). Value type is **string**.
++ `token_query_keyword` represents the parameter name used instead of "access_token" in the query of an API request according to [RFC 6750](http://tools.ietf.org/html/rfc6750#section-2.3). Value type is **string**.
 + `example_token` represents an example bearer token value. Value type is **string**. "example_token" is the **default**.
 
 ##### Example
 
     + token (Bearer Authentication Scheme)
-        + keyword: `token`
+        + `token_header_keyword`: `token`
 
 ---
 
@@ -628,18 +631,21 @@ Defined by the `Authenticated` keyword in Markdown list entity followed by optio
 
     + Authenticated: <list>
 
-> **NOTE:** If name of an authentication scheme dervied from `OAuth2 Authentication Scheme` is used, it can optionally have comma seperated list of oauth scopes inside `[]` immediately following it.
+If name of an authentication scheme dervied from `OAuth2 Authentication Scheme` is used, it can optionally have comma seperated list of OAuth scopes inside `[]` immediately following it. If two or more OAuth scopes are using in an authentication setting, then any of the scopes can be used to satisfy the authentication setting.
+
+> **NOTE:** Since each of the OAuth scopes is an [MSON Value][], they should be escaped by backticks if they contain any [MSON][] reserved characters
 
 #### Description
 This section describes an authentication setting. Based on the parent section, the authentication setting being described is one of the following:
 
-1. Resource authentication setting ([Resource section](#def-resource-section))
-2. Action authentication setting ([Action section](#def-action-section))
-3. Request authentication setting ([Request section](#def-request-section))
+1. Resource group authentication setting ([Resource group section](#def-resource-group-section))
+2. Resource authentication setting ([Resource section](#def-resource-section))
+3. Action authentication setting ([Action section](#def-action-section))
+4. Request authentication setting ([Request section](#def-request-section))
 
 The authentication setting for a [Request section](#def-request-section) denotes the authentication needed to be sent along with the request to get the [Response section](#def-response-section) of the respective transaction example.
 
-If the authentication setting contains only the `Authenticated` keyword, the transaction example allows all the authentication schemes described in the [Authentication schemes section](#def-authenticationschemes-section).
+If the authentication setting contains only the `Authenticated` keyword, the transaction example allows all the authentication schemes described in the [Authentication Schemes section](#def-authentication-schemes-section).
 
 ```
 + Authenticated
@@ -654,6 +660,26 @@ If the authentcation setting contains a list of scheme names after the `Authenti
 ```
 + Authenticated: oauth[read, email], token
 ```
+
+#### Resource group Authenticated description
+Description of the resource group authentication setting.
+
+If defined in a [Resource group section](#def-resource-group-section), all the requests of the transaction examples under the resources of this resource group will inherit the authentication setting unless specified otherwise.
+
+##### Example
+
+    # Group Users
+
+    + Authenticated
+
+    ## User [/user]
+    ### Retrieve User [GET]
+
+    + Response 200
+
+            {
+              "username": "john"
+            }
 
 #### Resource Authenticated description
 Description of the resource authentication setting.
@@ -1491,7 +1517,7 @@ With `varone := 42`, `vartwo = hello`, `varthree = 1024` the expansion is `/path
 
 [MSON]: https://github.com/apiaryio/mson
 [MSON Named Types]: https://github.com/apiaryio/mson/blob/master/MSON%20Specification.md#22-named-types
-[MSON Property Member Declaration]: https://github.com/apiaryio/mson/blob/master/MSON%20Specification.md#32-property-member-declaration
+[MSON Property Member Type]: https://github.com/apiaryio/mson/blob/master/MSON%20Specification.md#231-property-member-types
 [MSON Type Definition]: https://github.com/apiaryio/mson/blob/master/MSON%20Specification.md#35-type-definition
 [MSON Property Name]: https://github.com/apiaryio/mson/blob/master/MSON%20Specification.md#321-property-name
 [MSON Value]: https://github.com/apiaryio/mson/blob/master/MSON%20Specification.md#341-value
--- a/API Blueprint Specification.md
+++ b/API Blueprint Specification.md
@@ -30,6 +30,8 @@ Version: 1A8
 ### Section Basics
 + [Metadata section](#def-metadata-section)
 + [API name & overview section](#def-api-name-section)
++ [Authentication section](#def-authentication-section)
++ [Authentication schemes section](#def-authenticationschemes-section)
 + [Resource group section](#def-resourcegroup-section)
 + [Resource section](#def-resource-section)
 + [Resource model section](#def-model-section)
@@ -79,6 +81,9 @@ All of the blueprint sections are optional. However, when present, a section **m
 
 + [`0-1` **Metadata** section](#def-metadata-section)
 + [`0-1` **API Name & overview** section](#def-api-name-section)
++ [`0-1` **Authentication** section](#def-authentication-section)
+    + [`1` **Authentication schemes** section](#def-authenticationschemes-section)
+    + [`1` **Response** section](#def-response-section)
 + [`0+` **Resource** sections](#def-resource-section)
     + [`0-1` **URI Parameters** section](#def-uriparameters-section)
     + [`0-1` **Attributes** section](#def-attributes-section)
@@ -89,6 +94,7 @@ All of the blueprint sections are optional. However, when present, a section **m
         + [`0-1` **Schema** section](#def-schema-section)
     + [`1+` **Action** sections](#def-action-section)
         + [`0-1` **Relation** section](#def-relation-section)
+        + [`0-1` **Authentication** section](#def-authentication-section)
         + [`0-1` **URI Parameters** section](#def-uriparameters-section)
         + [`0-1` **Attributes** section](#def-attributes-section)
         + [`0+` **Request** sections](#def-request-section)
@@ -200,6 +206,7 @@ Following reserved keywords are used in section definitions:
 
 #### Header keywords
 - `Group`
+- `Authentication`
 - `Data Structures`
 - [HTTP methods][httpmethods] (e.g. `GET, POST, PUT, DELETE`...)
 - [URI templates][uritemplate] (e.g. `/resource/{id}`)
@@ -216,6 +223,8 @@ Following reserved keywords are used in section definitions:
 - `Values`
 - `Attribute` & `Attributes`
 - `Relation`
+- `Authentication`
+- `Authentication Schemes`
 
 > **NOTE: Avoid using these keywords in other Markdown headers or lists**
 
@@ -475,6 +484,149 @@ Name and description of the API
 
     # Basic ACME Blog API
     Welcome to the **ACME Blog** API. This API provides access to the **ACME Blog** service.
+
+---
+
+<a name="def-authentication-section"></a>
+## Authentication section
+- **Parent sections:** none, [Action section](#def-action-section)
+- **Nested sections:** [`1` Authentication schemes section](#def-authenticationschemes-section), [`1` Response section](#def-response-section)
+- **Markdown entity:** header, list
+- **Inherits from**: none
+
+#### Definition
+Defined by the `Authentication` keyword in Markdown header entity.
+
+    # Authentication
+
+**-- or --**
+
+Defined by the `Authentication` keyword in Markdown list entity followed by optional comma seperated list of scheme names which is prefixed by a colon.
+
+    + Authentication: <list>
+
+> **NOTE:** The second notation is only used when the parent section is an [Action section](#def-action-section).
+
+#### Description
+Based on the parent section, the usage of this section differs as shown below.
+
+#### API Authentication section
+Description of the authentication support for the whole API.
+
+This section **should** include one [Authentication schemes section](#def-authenticationschemes-section) and one [Response section](#def-response-section).
+
+The [Authentication schemes section](#def-authenticationschemes-section) defines all the types of authentication schemes supported by the API. Where as, the [Response section](#def-response-section) describes the error response given by an authentication protected endpoint when no authentication is provided with the request.
+
+##### Example
+
+    ## Authentication
+    This API supports different kinds of the authentication.
+
+    + Authentication Schemes
+        + ...
+
+    + Response 403 (application/json)
+
+            {
+              "message": "Authentication required",
+              "type": "error"
+            }
+
+#### Action Authentication section
+Description of the authentication support for a particular action in the API.
+
+If defined, the respecitve action of the [Action section](#def-action-section) in the API can only be accessed using a vaild authentication scheme. If a list of scheme names is provided after the `Authentication` keyword, the respective action in the API does not allow other authentication schemes.
+
+This section **should** not have any nested sections.
+
+##### Example
+
+```
++ Authentication
+```
+
+```
++ Authentication: basic, token
+```
+
+---
+
+<a name="def-authenticationschemes-section"></a>
+## Authentication schemes section
+- **Parent sections:** [Authentication section](#def-authentication-section)
+- **Nested sections:** See below
+- **Markdown entity:** list
+- **Inherits from**: none
+
+#### Definition
+Defined by the `Authentication Schemes` keyword.
+
+    + Authentication Schemes
+
+#### Description
+Description of the list of authentication schemes supported by the API.
+
+This section **must** be composed of nested list items only. This section **must not** contain any other elements. Each list item describes a single Authentication scheme.
+
+    + <scheme name> (<base scheme type>) - <description>
+
+        + `<key 1>`: <value 1>
+        + `<key 2>`: <value 2>
+        ...
+        + `<key N>`: <value N>
+
+Where:
+
++ `<scheme name>` is the scheme name as written in [Authentication section](#def-authentication-section) under an [Action section](#def-action-section).
++ `<description>` is any **optional** Markdown-formatted description of the authentication scheme.
++ `<base scheme type>` is the authentication scheme type as expected by the API (e.g. "Basic Auth Scheme"). See below for all the available base authentication schemes.
++ `<key n>` represents a key of an customizable option in the base authentication scheme.
++ `<value n>` represents the value for the option with the key `<key n>`.
+
+Description of the base authentication schemes inherently defined in API Blueprint are defined below.
+
+#### Basic Auth Scheme
+This is the [Basic Authentication Scheme](http://tools.ietf.org/html/rfc2617#section-2) as specified in [RFC 2617](http://tools.ietf.org/html/rfc2617).
+
+The customizable options for this scheme are:
+
++ `sample_username` represents a sample username value. "sample_user" is the **default**.
++ `sample_password` represents a sample password value. "sample_pass" is the **default**.
+
+##### Example
+
+    + basic (Basic Auth Scheme)
+        + sample_username: `bond`
+        + sample_password: `Iam007`
+
+#### Oauth2 Auth Scheme
+This scheme uses the [Oauth2 Authorization Framework](http://tools.ietf.org/html/rfc6749) specified in [RFC 6749](http://tools.ietf.org/html/rfc6749) to get an authorization grant and uses it access the API.
+
+The customizable options for this scheme are:
+
++ `authorization_endpoint` is the authorization endpoint required for the authorization grant. "/authorize" is the **default**.
++ `access_token_endpoint` is the token endpoint used by client to obtain the access token. "/access_token" is the **default**.
++ `token_header_keyword` represents the string used instead of "Bearer" in "Authorization" header field according to [RFC 6750](http://tools.ietf.org/html/rfc6750#section-2.1). "Bearer" is the **default**.
++ `token_body_keyword` represents the parameter name used instead of "access_token" in the body of an API request according to [RFC 6750](http://tools.ietf.org/html/rfc6750#section-2.2). "access_token" is the **default**.
++ `token_query_keyword` represents the parameter name used instead of "access_token" in the query of an API request according to [RFC 6750](http://tools.ietf.org/html/rfc6750#section-2.3). "access_token" is the **default**.
+
+##### Example
+
+    + oauth (Oauth2 Auth Scheme)
+        + access_token_endpoint: `/token`
+        + token_header_keyword: `token`
+
+#### Bearer Auth Scheme
+This scheme is popularly used when API consumers has access to their own unique access token which can be used as a bearer token as specified in [RFC 6750](https://tools.ietf.org/html/rfc6750#section-2.1).
+
+The customizable options for this scheme are:
+
++ `keyword` represents the string used instead of "Bearer" in the "Authorization" header field according to [RFC 6750](https://tools.ietf.org/html/rfc6750#section-2.1). "Bearer" is the **default**.
+
+##### Example
+
+    + token (Bearer Auth Scheme)
+        + keyword: `token`
 
 ---
 
@@ -768,7 +920,7 @@ One HTTP request-message example – payload.
 <a name="def-response-section"></a>
 ## Response section
 - **Abstract**
-- **Parent sections:** [Action section](#def-action-section)
+- **Parent sections:** [Action section](#def-action-section) | [Authentication section](#def-authentication-section)
 - **Nested sections:** [Refer to payload section](#def-payload-section)
 - **Markdown entity:** list
 - **Inherits from**: [Payload section](#def-payload-section)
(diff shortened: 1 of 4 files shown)


pksunkara:
Authentication proposal is ready. @Almad @zdne @kylef @danielgtaylor @klokane

API Blueprint Specification.md
```diff
@@ -478,6 +487,149 @@ Name and description of the API
...
++ `access_token_endpoint` is the token endpoint used by client to obtain the access token. "/access_token" is the **default**.
++ `token_header_keyword` represents the string used instead of "Bearer" in "Authorization" header field according to [RFC 6750](http://tools.ietf.org/html/rfc6750#section-2.1). "Bearer" is the **default**.
++ `token_body_keyword` represents the parameter name used instead of "access_token" in the body of an API request according to [RFC 6750](http://tools.ietf.org/html/rfc6750#section-2.2). "access_token" is the **default**.
++ `token_query_keyword` represents the parameter name used instead of "access_token" in the query of an API request according to [RFC 6750](http://tools.ietf.org/html/rfc6750#section-2.3). "access_token" is the **default**.
+
+##### Example
+
+    + oauth (Oauth2 Auth Scheme)
+        + access_token_endpoint: `/token`
+        + token_header_keyword: `token`
```
danielgtaylor: It might make sense to add `sample_token` with a default of something sane, like `c7b216282541` (e.g. string of random hex chars, but be consistent as people may check generated documentation into version control). Same goes for the bearer auth scheme below.

API Blueprint Specification.md
```diff
@@ -478,6 +487,149 @@ Name and description of the API
...
+
+            {
+              "message": "Authentication required",
+              "type": "error"
+            }
+
+#### Action Authentication section
+Description of the authentication support for a particular action in the API.
+
+If defined, the respecitve action of the [Action section](#def-action-section) in the API can only be accessed using a vaild authentication scheme. If a list of scheme names is provided after the `Authentication` keyword, the respective action in the API does not allow other authentication schemes.
```
danielgtaylor: What about resources which provide public information without authentication *and* more information when authenticated (and maybe when the authenticated user is the owner)? For example:

```http
GET /users HTTP/1.1
```
```http
HTTP/1.1 200 OK
Content-Type: application/json

{
  "users": [
    {
      "name": "Daniel"
    }
  ]
}
```

vs.

```http
GET /users HTTP/1.1
Authorization: Bearer abc123
```
```http
HTTP/1.1 200 OK
Content-Type: application/json

{
  "users": [
    {
      "name": "Daniel",
      "email": "daniel@example.com",
      "phone": "(206) 123-45678"
    }
  ]
}
```

Do you think it complicates the specification too much to add `+ Authentication (optional)` and an `+ Authenticated` flag for a request or something similar? The transaction examples would likely help to explain the exact details. Example:

```apib
# Foo Resource [/foo]

## Get a foo [GET]

+ Authentication (optional): oauth2, bearer

+ Request

+ Response 200 (application/json)

    + Attributes
        + users (array, required)
            + (object)
                + name

+ Request

    + Authenticated

+ Response 200 (application/json)

    + Attributes
        + users (array, required)
            + (object)
                + name
                + email
                + phone
```

danielgtaylor:
I really like what you've proposed here and I think this is a great start to supporting a feature that I've long wished was built into API Blueprint.

One thing that I keep coming back to is the list of base scheme types: `basic`, `oauth2`, and `bearer` (which is a subset of OAuth 2). I feel like we may need to add another base type which allows you to define headers/query params/etc with no built-in logic. For example, I should be able to say that I have an API where every request must contain the query parameter `api_key` and a `app_id`. None of the existing schemes allow this to work. If instead you allow arbitrary headers and query parameters then Apiary, Aglio, Dredd and other tools could make use of this in the UI, for example letting the user set this once then making multiple example requests. You can then easily make the `bearer` type using this generic base (though we'd probably still provide it as a built-in due to its popularity):

```apib
## Authentication

+ Authentication Schemes

    + myauth (Generic Base)

        My custom authentication scheme. You must provide an API key, which is a unique
        identifier for your account, as well as an application identifier that helps to
        track which of your applications is making calls to the API. You can find
        these in your [developer dashboard](...).

        + Query Parameters
            + api_key: abcd1234
            + app_id: fefe8989

    + bearer (Generic Base)
        + Headers
            + Authentication: Bearer abcd1234

## Frob [/frob/{id}]
### Get a frob [GET]
+ Authentication: myauth
...
```

We need some way to say what the header/query name is and a sample value. In Aglio, for example, this could be exposed by a new authentication section before the groups/resources start. Each action could have a tool to make a sample request, and you would be asked to fill in the `api_key` and `app_id` only once, leaving more space for other parameters/bodies of each individual action. Thoughts?

API Blueprint Specification.md
pksunkara: Yeah, I meant to do that but kinda forgot it :disappointed:

I think defaulting the `sample_token` to `sample_token` should not be a problem, right? Because Oauth spec does not specify any rule for the access tokens.

API Blueprint Specification.md
pksunkara: A solution to this would be to emulate how `Attributes` is done currently. Allow `Authentication` section to be in either `Action` section or a `Request` section. If it is in `Action` section, it will be default for all the child requests while allowing an `Authentication` section in the request to override it.

We don't need to have a new keyword `Authenticated` or add `optional` suffixes. What do you say?

pksunkara:
> I feel like we may need to add another base type which allows you to define headers/query params/etc with no built-in logic.

> You can then easily make the bearer type using this generic base

That was my thought process too. Let me explain it.

The concept behind the current syntax proposal is that an API Blueprint writer can develop his own custom authentication methods using MSON (using some kinda function/lambda syntax). But since MSON is not yet that mature and this is only the first iteration (basic support) of authentication, I have decided to add only 3 `Base Auth Schemes`.

In background, all these 3 auth schemes are doing conceptually nothing but extending the headers and query params of the request. An example of Bearer Auth Scheme can be seen below. (Pay attention to the template kind of stuff).

```
## Bearer Auth Scheme (Auth Scheme)
+ Headers
    + Authorization: *keyword* *sample_token*
```

I haven't proposed the custom authentication schemes because I felt that we would not do justice for it with the current state of MSON (even if we disregard functions) and whatever we implement is going to be very volatile due to it. We do not want the users to start using it and turn out we need to change how it works.

So, I have decided to propose only the 3 authentication schemes (which will be defined in parser) and consider this the first iteration of authentication rollout. I would love to do the second iteration later where these schemes are actually written in API Blueprint and parser just parses it. But as I said, we need to give some time for it.

Please ask me if you don't understand something I said.

API Blueprint Specification.md
fosrias: I think that this scenario makes perfect sense in the request samples sense as @pksunkara suggests. If you want users to know what happens nothing is clearer than a sample. However, our experience of multiple samples and the ability to switch between them semantically in the machine column is non-existent ATM. So, not sure how the user would see this in the docs unfortunately, even though I think it is the simpler solution.

API Blueprint Specification.md
```diff
@@ -478,6 +487,149 @@ Name and description of the API
...
++ `token_body_keyword` represents the parameter name used instead of "access_token" in the body of an API request according to [RFC 6750](http://tools.ietf.org/html/rfc6750#section-2.2). "access_token" is the **default**.
++ `token_query_keyword` represents the parameter name used instead of "access_token" in the query of an API request according to [RFC 6750](http://tools.ietf.org/html/rfc6750#section-2.3). "access_token" is the **default**.
+
+##### Example
+
+    + oauth (Oauth2 Auth Scheme)
+        + access_token_endpoint: `/token`
+        + token_header_keyword: `token`
+
+#### Bearer Auth Scheme
```
fosrias: Suggested: Bearer Authentication Scheme. Should not use shortened names in public document. Whole words. Here and elsewhere if applicable.

zdne: :+1:

API Blueprint Specification.md
pksunkara: Yes, as @fosrias said, I am not sure about how the authentication would be documented with Transaction examples. On the other hand, documenting it at the action will be pretty easy.

Hmm.. I am now thinking if it is better if we do not have a `Request Authentication` and just leave the `Action Authentication` as it is.

API Blueprint Specification.md
pksunkara: @fosrias Are you talking about the headers and description in API Blueprint spec markdown? Or are you talking about the actual type name (e.g. `Basic Auth Scheme` etc..) in API Blueprint?

API Blueprint Specification.md
danielgtaylor: Right, I think that `sample_token` will work fine :+1:

API Blueprint Specification.md
danielgtaylor: I think your original idea of just adding `+ Authentication: oauth2` inside of the request is okay by me. I should be able to handle that easily enough in Aglio and I'm sure the Apiary docs can be updated to do so as well.

danielgtaylor:
@pksunkara your explanation of how things will work in the future with MSON makes sense. As long as that is on the radar, then I'm good with shipping these three base auth schemes, since I think it probably covers 80% of APIs.

API Blueprint Specification.md
fosrias: Well, I don't know about the type name so, probably the former in the spec documentation @pksunkara. That section and the basic section are descriptions. So, basically change it inline where I made the comment.

pksunkara:
Updated with Request Authentication section. I tried to describe as much as I can. Please propose any wording changes if you want. Thanks.

danielgtaylor:
@pksunkara I think this makes sense in the latest commit. Thanks for adding this!

API Blueprint Specification.md
```diff
@@ -30,6 +30,8 @@ Version: 1A8
 ### Section Basics
 + [Metadata section](#def-metadata-section)
 + [API name & overview section](#def-api-name-section)
++ [Authentication section](#def-authentication-section)
```
zdne: Similarly to Data Structures I would like to move these two sections from "Basics" to "Going Further" – if you are trying to get a grasp on what API Blueprint is the authentication sections are not the first things you should be learning..

API Blueprint Specification.md
```diff
@@ -478,6 +488,189 @@ Name and description of the API
...
+**-- or --**
+
+Defined by the `Authentication` keyword in Markdown list entity followed by optional comma seperated list of scheme names which is prefixed by a colon.
+
+    + Authentication: <list>
+
+> **NOTE:** The second notation is only used when the parent section is either an [Action section](#def-action-section) or a [Request section](#def-request-section).
+
+#### Description
+Based on the parent section, the usage of this section differs as shown below.
```
zdne: Overall description instead of just 

> Based on the parent section, the usage of this section differs as shown below.

Something along these lines: 

> This section does/represents XXXX

zdne: Trying to figure out what is the purpose of the section: Reading "authentication settings" vs. "authentication support" 

Consider what you have there: 

- Description of the authentication support for the whole API
- Description of the request level authentication settings
- Description of the action level authentication settings

Please stick with one (support vs settings) – I am in favor of the settings or perhaps another word (mechanism?)

API Blueprint Specification.md
```diff
@@ -478,6 +488,189 @@ Name and description of the API
...
+
+#### API Authentication section
```
zdne: I would suggest to follow the [attributes approach](https://github.com/apiaryio/api-blueprint/blob/pksunkara/auth/API%20Blueprint%20Specification.md#description-16) – explain as much as you can in the general description – then mention where it can be used and then get into context-specific details.

API Blueprint Specification.md
```diff
@@ -478,6 +488,189 @@ Name and description of the API
...
+        + ...
+
+    + Response 403 (application/json)
+
+            {
+              "message": "Authentication required",
+              "type": "error"
+            }
+
+#### Request Authentication section
```
zdne: Logical ordering – from top most to down – this should be after action

1. API Authentication section
2. Action Authentication section
3. Request Authentication section

API Blueprint Specification.md
```diff
@@ -478,6 +488,189 @@ Name and description of the API
...
+
+> **NOTE:** The second notation is only used when the parent section is either an [Action section](#def-action-section) or a [Request section](#def-request-section).
+
+#### Description
+Based on the parent section, the usage of this section differs as shown below.
+
+#### API Authentication section
+Description of the authentication support for the whole API.
+
+This section **should** include one [Authentication schemes section](#def-authenticationschemes-section) and one [Response section](#def-response-section).
```
zdne: I think having the limit for one response section a bit blunt – what about different media types based on content - negotiation? e.g application/xml vs application/xml etc. 

Any problem in making this working for multiple responses?

API Blueprint Specification.md
```diff
@@ -478,6 +488,189 @@ Name and description of the API
...
+
+The [Authentication schemes section](#def-authenticationschemes-section) defines all the types of authentication schemes supported by the API. Where as, the [Response section](#def-response-section) describes the error response given by an authentication protected endpoint when no authentication is provided with the request.
```
zdne: "Where as, the Response section.." sounds odd

API Blueprint Specification.md
pksunkara: Well, if the response types are going to be differed only in content-types, then, we should not have a problem.

API Blueprint Specification.md
zdne: describes the error message returned in the case of failed authentication...

API Blueprint Specification.md
pksunkara: I would agree but putting the Action Authentication section description above the Request Authentication section description in the spec wasn't looking good. What do you say?

API Blueprint Specification.md
```diff
@@ -478,6 +488,189 @@ Name and description of the API
...
+
+            {
+              "message": "Authentication required",
+              "type": "error"
+            }
+
+#### Request Authentication section
+Description of the request level authentication settings.
+
+If defined inside a [Request section](#def-request-section) of a HTTP transaction example, it implies that the API's response will be the same as the respective response when a valid authentication is sent along with the above mentioned request. This allows another HTTP transaction example to be specified for the cases where an authentication is not needed.
```
zdne: > If defined inside a [Request section](#def-request-section) of a HTTP transaction example, it implies that the API's response will be the same as the respective response when a valid authentication is sent along with the above mentioned request.

This is a super-long and complicated sentence. Please break it down. Also the first part is not needed as it is already implied by the heading "Request Authentication"

zdne: > the API's response will be the same as the respective response when a valid authentication is sent along with the above mentioned request.

No clue what this is saying

API Blueprint Specification.md
```diff
@@ -478,6 +488,189 @@ Name and description of the API
...
+                "username": "bond"
+            }
+
+    + Request
+        + Authentication
+
+    + Response 200 (application/json)
+
+            {
+                "username": "bond",
```
zdne: :gun: as much as I like 007 I would stick with the words / payloads already used throughout the specification

API Blueprint Specification.md
```diff
@@ -478,6 +488,189 @@ Name and description of the API
...
+    + Response 200
+
+            { ... }
+
+---
+
+<a name="def-authenticationschemes-section"></a>
+## Authentication schemes section
+- **Parent sections:** [Authentication section](#def-authentication-section)
+- **Nested sections:** See below
```
zdne: If I understand it correctly, there are no nested sections. The content of this section (key-pairs) aren't sections...

pksunkara: Yes, they are conceptually MSON but not actually an MSON section.

API Blueprint Specification.md
```diff
@@ -478,6 +488,189 @@ Name and description of the API
...
+            }
+
+#### Request Authentication section
+Description of the request level authentication settings.
+
+If defined inside a [Request section](#def-request-section) of a HTTP transaction example, it implies that the API's response will be the same as the respective response when a valid authentication is sent along with the above mentioned request. This allows another HTTP transaction example to be specified for the cases where an authentication is not needed.
+
+This section **should** not have any nested sections.
+
+If a list of scheme names is provided after the `Authentication` keyword, then only those particular authentication schemes will be allowed for the respective HTTP transaction.
```
zdne: Where is this explained / defined. I feel you are talking about something that has not been defined.

API Blueprint Specification.md
```diff
@@ -478,6 +488,189 @@ Name and description of the API
...
+- **Markdown entity:** list
+- **Inherits from**: none
+
+#### Definition
+Defined by the `Authentication Schemes` keyword.
+
+    + Authentication Schemes
+
+#### Description
+Description of the list of authentication schemes supported by the API.
```
zdne: Skip the "Description" word

zdne: Also I miss explanation of what "authentication scheme" is. Please add it

zdne: I feel there is a whole missing explanation of schemes and scheme inheritance

API Blueprint Specification.md
```diff
@@ -478,6 +488,189 @@ Name and description of the API
...
+    + <scheme name> (<base scheme type>) - <description>
+
+        + `<key 1>`: <value 1>
+        + `<key 2>`: <value 2>
+        ...
+        + `<key N>`: <value N>
+
+Where:
+
++ `<scheme name>` is the scheme name as written in [Authentication section](#def-authentication-section) under an [Action section](#def-action-section).
```
zdne: so the scheme is defined here or in the Authentication section

API Blueprint Specification.md
```diff
@@ -478,6 +488,189 @@ Name and description of the API
...
+
+The customizable options for this scheme are:
+
++ `sample_username` represents a sample username value. "sample_user" is the **default**.
++ `sample_password` represents a sample password value. "sample_pass" is the **default**.
+
+##### Example
+
+    + basic (Basic Auth Scheme)
+        + sample_username: `bond`
```
zdne: again, consider using sample values matching the theme of examples in the specification

API Blueprint Specification.md
```diff
@@ -478,6 +488,189 @@ Name and description of the API
...
++ `<value n>` represents the value for the option with the key `<key n>`.
+
+Description of the base authentication schemes inherently defined in API Blueprint are defined below.
+
+#### Basic Authentication Scheme
+This is the [Basic Authentication Scheme](http://tools.ietf.org/html/rfc2617#section-2) as specified in [RFC 2617](http://tools.ietf.org/html/rfc2617).
+
+The customizable options for this scheme are:
+
++ `sample_username` represents a sample username value. "sample_user" is the **default**.
```
zdne: Capitalization. Consider using `-` as with MSON inline description

API Blueprint Specification.md
```diff
@@ -478,6 +488,189 @@ Name and description of the API
...
++ `sample_password` represents a sample password value. "sample_pass" is the **default**.
+
+##### Example
+
+    + basic (Basic Auth Scheme)
+        + sample_username: `bond`
+        + sample_password: `Iam007`
+
+#### Oauth2 Authentication Scheme
+This scheme uses the [Oauth2 Authorization Framework](http://tools.ietf.org/html/rfc6749) specified in [RFC 6749](http://tools.ietf.org/html/rfc6749) to get an authorization grant and uses it access the API.
```
zdne: Uses 2x in one sentence. Break it down.

API Blueprint Specification.md
```diff
@@ -478,6 +488,189 @@ Name and description of the API
...
+        + `<key 2>`: <value 2>
+        ...
+        + `<key N>`: <value N>
+
+Where:
+
++ `<scheme name>` is the scheme name as written in [Authentication section](#def-authentication-section) under an [Action section](#def-action-section).
++ `<description>` is any **optional** Markdown-formatted description of the authentication scheme.
++ `<base scheme type>` is the authentication scheme type as expected by the API (e.g. "Basic Auth Scheme"). See below for all the available base authentication schemes.
++ `<key n>` represents a key of an customizable option in the base authentication scheme.
```
zdne: you have just introduced "customizable options" – what are they where were they defined? I believe these terms should be mentioned early in the description above and this breakdown should just "reference" the term

API Blueprint Specification.md
```diff
@@ -478,6 +488,189 @@ Name and description of the API
...
++ `sample_token` represens a sample access token value. "sample_token" is the **default**.
+
+##### Example
+
+    + oauth (Oauth2 Auth Scheme)
+        + access_token_endpoint: `/token`
+        + token_header_keyword: `token`
+
+#### Bearer Authentication Scheme
+This scheme is popularly used when API consumers has access to their own unique access token which can be used as a bearer token as specified in [RFC 6750](https://tools.ietf.org/html/rfc6750#section-2.1).
```
zdne: > This scheme is popularly used when API consumers has access to their own unique access token which can ...

There is no room for justification in the spec. No need to justify ourselves. Throw this part out, keep it simple.

API Blueprint Specification.md
```diff
@@ -1034,7 +1228,7 @@ Defined by the `Data Structures` keyword.
 #### Description
 This section holds arbitrary data structures definitions defined in the form of [MSON Named Types][].
 
-Data structures defined in this section **may** be used in any [Attributes section][]. Similarly, any data structures defined in a [Attributes section][] of a named [Resource Section][] **may** be used in a data structure definition.
+Data structures defined in this section **may** be used in any [Attributes section][]. Similarly, any data structures defined in a [Attributes section][] of a named [Resource section][] **may** be used in a data structure definition.
```
zdne: As tempting as it is, please avoid doing fixes to other things in this PR – it will be easier to review and maintain. Make another PR to fix things like these.

examples/13. Authentication.md
```diff
@@ -0,0 +1,57 @@
+FORMAT: 1A
+
+# Authentication API
+
+## API Blueprint
++ [Previous: Advanced Action](12.%20Advanced%20Action.md)
++ [This: Raw API Blueprint](https://raw.github.com/apiaryio/api-blueprint/master/examples/13.%20Authentication.md)
+
+## Authentication
+The allowed authentication schemes for the API server are defined below.
```
zdne: API server sounds odd

examples/13. Authentication.md
```diff
@@ -0,0 +1,57 @@
...
+
+### Retrieve a User Profile [GET]
+This endpoint needs authentication from the user before retrieving his profile data.
+
++ Authentication
+
++ Response 200 (application/json)
+
+        {
+          "username": "bond",
```
zdne: Check what payloads are used throughout other examples. I think it is worth to try to keep this unified. That way it will be less confusing for those user that are browsing through examples as whole.

zdne:
@pksunkara Good job! Overall I think the concept is pretty solid. This PR needs however some thinking about formulating the ideas into words – I found the descriptions of the sections a little bit confusing. Often discussing things that are not need (e.g. why we are doing it).

As for the proposed functionality itself I have two main concerns: 

1. Limiting the "error" response to just one payload 
2. As @danielgtaylor pointed out – I would like to see an example where calling an endpoint authenticated returns one (200) response and calling the same endpoint not authenticated returns another (200) response..

Maybe consider breaking the examples into two:

1. Simple basic auth on the whole API
2. Advanced basic auth + Oauth

zdne:
@pksunkara as for the descriptions – try to identify the main concepts: 

- authentication definition
- schemes
- scheme options

and focus on explaining those

pksunkara:
@zdne Thanks for the review. I will work on fixing them. My main question is,

> Limiting the "error" response to just one payload

I understand that a user might want to showcase different responses based on the different media types. Is that enough here? Or do we need to accommodate other kinds of responses?

API Blueprint Specification.md
kylef: I think we should call this example instead of sample, to be consistent with MSON which uses the term example.

screen shot 2015-06-04 at 08 46 53
screen shot 2015-06-04 at 08 46 57

API Blueprint Specification.md
zdne: Not sure what you mean by "looking good". While esthetic is important I would argue that in the case of a formal spec the logical ordering is even more important.

pksunkara: Sorry, When I mean "looking good", I meant, logical order. Since Action Auth Section is nothing defaults for Request Auth Section, I wanted to introduce Request Auth Section first.

API Blueprint Specification.md
zdne: It is also quite unclear from the spec what happens in the case where you want to return something else for unauthenticated vs authenticated request. Is this response used every time an unauthenticated request is made? 

I think there is potentially a lot of edge cases where this behavior may be limiting.

API Blueprint Specification.md
zdne: Well my point is the nested section should say "none"

API Blueprint Specification.md
zdne: Agree with Kyle. Maybe also disregard my comment on using `-` as we are not using it in the rest of the spec.

zdne:
> I understand that a user might want to showcase different responses based on the different media types. Is that enough here? Or do we need to accommodate other kinds of responses?

As per my related comment https://github.com/apiaryio/api-blueprint/pull/201#discussion_r31800156 I think this is a potentially limiting and confusing. I like the idea of providing a default error response. But there should be a simple and clear way to provide non-error responses to requests without authentication..

API Blueprint Specification.md
```diff
@@ -478,6 +488,189 @@ Name and description of the API
...
+
+##### Example
+
+    ## Retrieve User [GET]
+
+    + Authentication
+
+    + Response 200
+
+            { ... }
```
Almad: All other examples are valid JSONs, I'd preserve it that way.

API Blueprint Specification.md
```diff
@@ -478,6 +488,189 @@ Name and description of the API
...
+Description of the request level authentication settings.
+
+If defined inside a [Request section](#def-request-section) of a HTTP transaction example, it implies that the API's response will be the same as the respective response when a valid authentication is sent along with the above mentioned request. This allows another HTTP transaction example to be specified for the cases where an authentication is not needed.
+
+This section **should** not have any nested sections.
+
+If a list of scheme names is provided after the `Authentication` keyword, then only those particular authentication schemes will be allowed for the respective HTTP transaction.
+
+```
++ Authentication
```
Almad: If not specified, wouldn't it be more readable to add just `Authenticated` instead?

pksunkara: It would involve an additional reserved keyword. We don't want this to be too complex.

API Blueprint Specification.md
```diff
@@ -79,6 +81,9 @@ All of the blueprint sections are optional. However, when present, a section **m
 
 + [`0-1` **Metadata** section](#def-metadata-section)
 + [`0-1` **API Name & overview** section](#def-api-name-section)
++ [`0-1` **Authentication** section](#def-authentication-section)
+    + [`1` **Authentication schemes** section](#def-authenticationschemes-section)
+    + [`1` **Response** section](#def-response-section)
 + [`0+` **Resource** sections](#def-resource-section)
```
Almad: Shouldn't it be possible to specify `Authentication` on `Resource` section as well?

Almad:
As a non-professional RFCer, I have stupid questions.

1) Ad example, I have no clue what `description` in `attributes` will propagate to.  Is it description of the given authentication? Would it propagate into authentication/payload?

If it is MSON approach to describe payload, I don't know why request and response are inconsistent...and actually, why it's inconsistent across the whole example. 

2) How do I describe OAuth2 scopes — in Attributes of the quthentication? No clue about it.

3) How do I override oauth scopes for particular Resource or Action? 

4) Can I set `oauth_resource_uri` for oauth2?

5) YMMV, but would vote for `oauth2`. I think this can bite us.

6) I know we want to get rid of `HOST`, but in the same fashion, I'd currently add a way to specify default `authorize_url` (or `_hostname`) until we have a better notion of specifying deployment targets; as far as I observed, authentication server usually live on a different domain/namespace then the API itself. 

if the `_endpoint` are meant to allow inclusion of hostname, I'd add that to spec. 

7) Thinking whether there shouldn't be a way to bind a response to particular authentication scheme, i.e. `401` response to `basic` only, but not to `oauth`. Not sure about this one. 

8) I'd add an example on how to return a different response based on whether request was authenticated. 

9) Not sure whether I can mark whole API as being behind default authentication, would be good to clarify.

timb-pt:
I'm dealing with a custom authentication scheme that requires a SSO token parameter in either the query string or a custom header.

There are two more optional (authentication-related) parameters that can be added to the query string, and they're not username, password, or another token; they're custom to our business case.

How can I represent our setup with this proposed authentication spec, or what changes need to be made so I can represent these extra parameters in a DRY manner?

The business case is currently limited in a manner that prevents practical integration of username/password, OAuth, or Bearer authentication.

API Blueprint Specification.md
```diff
@@ -478,6 +488,189 @@ Name and description of the API
...
+This is the [Basic Authentication Scheme](http://tools.ietf.org/html/rfc2617#section-2) as specified in [RFC 2617](http://tools.ietf.org/html/rfc2617).
+
+The customizable options for this scheme are:
+
++ `sample_username` represents a sample username value. "sample_user" is the **default**.
++ `sample_password` represents a sample password value. "sample_pass" is the **default**.
+
+##### Example
+
+    + basic (Basic Auth Scheme)
```
honzajavorek: `Basic Auth Scheme` or `Basic Authentication Scheme`? or is it alias? unclear

API Blueprint Specification.md
```diff
@@ -478,6 +488,189 @@ Name and description of the API
...
+        + sample_username: `bond`
+        + sample_password: `Iam007`
+
+#### Oauth2 Authentication Scheme
```
honzajavorek: What is preferred spelling? Oauth2 or OAuth2?

API Blueprint Specification.md
```diff
@@ -478,6 +488,189 @@ Name and description of the API
...
++ `authorization_endpoint` is the authorization endpoint required for the authorization grant. "/authorize" is the **default**.
++ `access_token_endpoint` is the token endpoint used by client to obtain the access token. "/access_token" is the **default**.
++ `token_header_keyword` represents the string used instead of "Bearer" in "Authorization" header field according to [RFC 6750](http://tools.ietf.org/html/rfc6750#section-2.1). "Bearer" is the **default**.
++ `token_body_keyword` represents the parameter name used instead of "access_token" in the body of an API request according to [RFC 6750](http://tools.ietf.org/html/rfc6750#section-2.2). "access_token" is the **default**.
++ `token_query_keyword` represents the parameter name used instead of "access_token" in the query of an API request according to [RFC 6750](http://tools.ietf.org/html/rfc6750#section-2.3). "access_token" is the **default**.
++ `sample_token` represens a sample access token value. "sample_token" is the **default**.
+
+##### Example
+
+    + oauth (Oauth2 Auth Scheme)
```
honzajavorek: again, Auth vs. Authentication mismatch

API Blueprint Specification.md
honzajavorek: if not thrown away, then: `API consumers has` -> `API consumers have`?

API Blueprint Specification.md
netmilk: Missing dash in link `#def-authenticationschemes-section`

API Blueprint Specification.md
```diff
@@ -478,6 +488,189 @@ Name and description of the API
...
+
+    + Authentication
+
+    + Response 200
+
+            { ... }
+
+---
+
+<a name="def-authenticationschemes-section"></a>
```
netmilk: NIT: I know it will work, but isn't a dash missing in the name?

netmilk:
HI,

- Form the API consumer standpoint: I'm afraid this proposed syntax is missing possibility to customize request (custom and templatable URIs, parameters and request bodies) for each step of oAuth dance and **can't be used for configuration of clients** as it is. Please have a look at [outhd authentication specification](https://github.com/oauth-io/oauthd/wiki/Specs%3A-Provider%27s-configuration) and their [gallery of real world API authentication configurations](https://github.com/oauth-io/oauthd/tree/master/providers). I dumped it into [one gist](https://gist.github.com/netmilk/e39df6061fe741f9ae71) for better browsability.

- From the API testing standpoint: Some action or resource theoretically doesn't need to be accessible when user is authenticated. If I'm not missing something it's not possible to mark action or resource not accessible of authenticated user.

- I'm missing syntax for revocation URI

- I'm missing syntax for token refresh URI

- I'm missing oAuth1 authentication scheme

- I'm completely missing syntax for scopes but it's probably planed for some future Authorization iteration

- One formal: Maybe it's only me, but It's unclear where are these "base authentication schemes" names defined in the proposed specification. They should be exactly listed imho.

- One question: Can I define absolutely custom authentication scheme not inherited from the base ones?

API Blueprint Specification.md
pksunkara: This the style which the spec is using before. I didn't want to change it.

pksunkara:
Addressing @almad's feedback

> 1) Ad example, I have no clue what description in attributes will propagate to. Is it description of the given authentication? Would it propagate into authentication/payload?

If you look at the example, the `Attributes` are nested under `Request` which makes them the payload body of the request. I don't know why you thought it was related to the authentication.

> If it is MSON approach to describe payload, I don't know why request and response are inconsistent...and actually, why it's inconsistent across the whole example.

Well, there should be no problem using them like that. Now, I think about it, yeah, it might be confusing. I will change them.

> 2) How do I describe OAuth2 scopes — in Attributes of the quthentication? No clue about it.

> 3) How do I override oauth scopes for particular Resource or Action?

I didn't intend to include OAuth2 scopes in the first iteration of the authentication implementation. It was planned for the second iteration. But if you can make a strong point to include it, I do not mind.

> 4) Can I set oauth_resource_uri for oauth2?

We can do that. This is something like `/oauth`, right?

> 6) I know we want to get rid of HOST, but in the same fashion, I'd currently add a way to specify default authorize_url (or _hostname) until we have a better notion of specifying deployment targets; as far as I observed, authentication server usually live on a different domain/namespace then the API itself. If the _endpoint are meant to allow inclusion of hostname, I'd add that to spec.

Yup. Will add that.

> 7) Thinking whether there shouldn't be a way to bind a response to particular authentication scheme, i.e. 401 response to basic only, but not to oauth. Not sure about this one.

Are you talking about different type of error responses for different authentication schemes? If you are talking that an endpoint should access `oauth` but not `basic`, then that is already covered.

> 8) I'd add an example on how to return a different response based on whether request was authenticated.

Will do.

> 9) Not sure whether I can mark whole API as being behind default authentication, would be good to clarify.

The problem is there is no section which is covering the whole API other than `API description section` and `Metadata section`. Putting it in `metadata` is the better solution but I didn't do it because I wanted to see if the users want something like that.

API Blueprint Specification.md
Almad: Too complex to what—write or implement?

Almad:
> > 1) Ad example, I have no clue what description in attributes will propagate to. Is it 
> description of the given authentication? Would it propagate into authentication/payload?  
>  
> If you look at the example, the `Attributes` are nested under `Request` which makes them  
> the payload body of the request. I don't know why you thought it was related to the authentication.  

I think I missed the nesting…or, as I’ve said elsewhere, the fact that response before this one is “pasted” and this particular one is MSON automatically triggered that something is special in my mind.

> > If it is MSON approach to describe payload, I don't know why request and response are  
> inconsistent...and actually, why it's inconsistent across the whole example.
>  
> Well, there should be no problem using them like that. Now, I think about it, yeah, it might  
> be confusing. I will change them.

Thanks (see above). 

> > 2) How do I describe OAuth2 scopes — in Attributes of the quthentication? No clue about  
> it.
>  
> > 3) How do I override oauth scopes for particular Resource or Action?
>  
> I didn't intend to include OAuth2 scopes in the first iteration of the authentication  
> implementation. It was planned for the second iteration. But if you can make a strong  
> point to include it, I do not mind.

Well, I am yet to see OAuth2 API _without_ scopes.


> > 4) Can I set oauth_resource_uri for oauth2?
>  
> We can do that. This is something like `/oauth`, right?

Yup, but it can be different (and is, I think in case of Microsoft’s servers). 

> > 7) Thinking whether there shouldn't be a way to bind a response to particular authentication  
> scheme, i.e. 401 response to basic only, but not to oauth. Not sure about this one.
>  
> Are you talking about different type of error responses for different authentication  
> schemes? If you are talking that an endpoint should access `oauth` but not `basic`, then  
> that is already covered.

Yes for the first part, not for the letter—I am talking about when you are accessing `basic` and don’t have permissions for that, `401` is returned.

When you do the same for `oauth`, `403` is returned instead. 


> > 9) Not sure whether I can mark whole API as being behind default authentication, would  
> be good to clarify.
>  
> The problem is there is no section which is covering the whole API other than `API description  
> section` and `Metadata section`. Putting it in `metadata` is the better solution but  
> I didn't do it because I wanted to see if the users want something like that.

Hmm, well, we are definitely already having those use-cases, see every API that uses current OAuth2 beta.

I also think that we should have section like this anyway as we are heading towards DRYness.

API Blueprint Specification.md
pksunkara: I have changed my opinion in the meanwhile and will be adding the `Authenticated` keyword.

pksunkara:
@Almad 

> Yes for the first part, not for the letter—I am talking about when you are accessing `basic` and don’t have permissions for that, `401` is returned. When you do the same for `oauth`, `403` is returned instead. 

I understand. I am going to see if we can do it easily.

> Hmm, well, we are definitely already having those use-cases, see every API that uses current OAuth2 beta.

Well, I can add an option to the metdata, but I still prefer to wait and see on this.

> I also think that we should have section like this anyway as we are heading towards DRYness.

Do you have any small example proposal?

pksunkara:
Addressing @netmilk's feedback,

> From the API consumer standpoint: I'm afraid this proposed syntax is missing possibility to customize request (custom and templatable URIs, parameters and request bodies) for each step of oAuth dance and can't be used for configuration of clients as it is. Please have a look at outhd authentication specification and their gallery of real world API authentication configurations. I dumped it into one gist for better browsability.

I did look at them when proposing the syntax and took care of almost all the customisations in OAuth2 spec. Do you have anything to specific to point here?

> From the API testing standpoint: Some action or resource theoretically doesn't need to be accessible when user is authenticated. If I'm not missing something it's not possible to mark action or resource not accessible of authenticated user.

Yeah, it is not possible to do this. Any public endpoint is always accessible even if the user is authenticated or not.

> I'm missing syntax for revocation URI
> I'm missing syntax for token refresh URI

Do you want these to be optionally defined by user too? I am okay with that and will add the options.

> I'm missing oAuth1 authentication scheme

We do not plan to support OAuth1 in this iteration.

> I'm completely missing syntax for scopes but it's probably planed for some future Authorization iteration

Scopes are going to be implemented now itself according to feedback by Lukas.

> One formal: Maybe it's only me, but It's unclear where are these "base authentication schemes" names defined in the proposed specification. They should be exactly listed imho.

Weren't they listed in the Authentication Schemes section? For ex: [here](https://github.com/apiaryio/api-blueprint/pull/201/files#diff-c4fbef240fc70f2374b2048d50827f80R627). I don't understand why it is unclear.

> One question: Can I define absolutely custom authentication scheme not inherited from the base ones?

Custom authentication schemes are planned for the future. Check this [comment](https://github.com/apiaryio/api-blueprint/pull/201#issuecomment-107874971) for my rationality behind it.

honzajavorek:
Just to be clear, I have no further objections or comments on this, especially above what have been already said. Nothing blocking. I have little time to follow the ongoing conversation with @Almad and @netmilk, but if I find the time, I will surely add a comment in case I have a thought to share.

pksunkara:
This PR is now updated and is ready to review.

API Blueprint Specification.md
```diff
@@ -478,6 +490,232 @@ Name and description of the API
...
+- **Nested sections:** none
+- **Markdown entity:** list
+- **Inherits from**: none
+
+#### Definition
+Defined by the `Authenticated` keyword in Markdown list entity followed by optional comma seperated list of scheme names which are prefixed by a colon.
+
+    + Authenticated: <list>
+
+> **NOTE:** If name of an authentication scheme dervied from `OAuth2 Authentication Scheme` is used, it can optionally have comma seperated list of oauth scopes inside `[]` immediately following it.
```
danielgtaylor: You may want to make this more prominent than just a side note. If you are using scopes then this is really important information.

Also, maybe we should mention that any given scope will have permissions (i.e. if you have `[scope1, scope2]` then you can use *either* `scope1` or `scope2` to access the resource). It wasn't immediately clear to me how it worked.

API Blueprint Specification.md
```diff
@@ -478,6 +490,232 @@ Name and description of the API
...
+        + <customizable option 2>
+        ...
+        + <customizable option N>
+
+Where:
+
++ `<scheme name>` is the authentication scheme name by which this scheme is represented in [Authenticated section](#def-authenticated-section).
++ `<description>` is any **optional** Markdown-formatted description of the authentication scheme.
++ `<base scheme type>` is the base authentication scheme type (e.g. "Basic Authentication Scheme"). See below for all the available base authentication schemes.
++ `<customizable option n>` represents a customizable option in the base authentication scheme. It is a [MSON Property Member Declaration][] without the [MSON Type Definition][] where the [MSON Property Name][] represents the name of the option and [MSON Value][] represents the value of the option.
```
honzajavorek: nitpick: `customizable option n` vs. `customizable option N` above

API Blueprint Specification.md
```diff
@@ -478,6 +490,232 @@ Name and description of the API
...
+    + basic (Basic Authentication Scheme)
+        + example_username: `john`
+        + example_password: `password`
+
+#### OAuth2 Authentication Scheme
+This scheme uses the [OAuth2 Authorization Framework](http://tools.ietf.org/html/rfc6749) specified in [RFC 6749](http://tools.ietf.org/html/rfc6749) to get an authorization grant which is then used to access the API.
+
+The customizable options for this scheme are:
+
++ `oauth_resource_endpoint` is the oauth resource endpoint where the oauth framework resides. Value type is **string**. "/" is the **default**.
```
honzajavorek:
```diff
- oauth framework
+ OAuth framework
```

API Blueprint Specification.md
```diff
@@ -478,6 +490,232 @@ Name and description of the API
...
+## Authentication section
+- **Parent sections:** none
+- **Nested sections:** [`1` Authentication schemes section](#def-authenticationschemes-section), [`1` Response section](#def-response-section)
+- **Markdown entity:** header
+- **Inherits from**: none
+
+#### Definition
+Defined by the `Authentication` keyword in Markdown header entity.
+
+    # Authentication
```
kylef: There should never be a case where this is level 1 heading and instead should be 2 (`##`) since it should sit within the API ([API Name section](https://github.com/apiaryio/api-blueprint/blob/master/API%20Blueprint%20Specification.md#api-name--overview-section)).

API Blueprint Specification.md
honzajavorek: :+1: to @danielgtaylor's comment. I vote for example and explanation of exact interpretation of how scopes are handled.

btw

```diff
- oauth scopes
+ OAuth scopes
```

API Blueprint Specification.md
honzajavorek: Technically, I think API Blueprint parser ignores levels of headers and by subsequent tooling they get deduced implicitly from the semantic hierarchy in AST. Not sure though. I think it works like that inside Apiary - it's always rendered the same regardless the actual level of the header.

Anyway, I agree it would be better to indicate a sub-header in the example not to confuse anyone.

API Blueprint Specification.md
```diff
@@ -478,6 +490,232 @@ Name and description of the API
...
+    ## Authentication
+    This API supports different kinds of the authentication.
+
+    + Authentication Schemes
+        + ...
+
+    + Response 403 (application/json)
+
+            {
+              "message": "Authentication required",
```
kylef: This message shouldn't be applicable to a 403 response. As per the HTTP specification ([rfc2616 section 10.4.4](http://www.ietf.org/rfc/rfc2616.txt)), where it states "authorization will **not** help".

> ### 10.4.4 403 Forbidden
> Authorization will not help and the request SHOULD NOT be repeated.

We should use a different example that doesn't go against RFC2616.

API Blueprint Specification.md
```diff
@@ -478,6 +490,232 @@ Name and description of the API
...
+              "type": "error"
+            }
+
+---
+
+<a name="def-authenticationschemes-section"></a>
+## Authentication schemes section
```
honzajavorek:
```diff
- schemes
+ Schemes
```

API Blueprint Specification.md
honzajavorek: The sad fact is that in real-world use this is, from my experience, deliberately violated just about everywhere. It's questionable whether to be RFC-purists or stick to what all the others do.

Seems like 401 could do the job - http://stackoverflow.com/a/6937030/325365

API Blueprint Specification.md
```diff
@@ -478,6 +490,232 @@ Name and description of the API
...
+        + example_username: `john`
+        + example_password: `password`
+
+#### OAuth2 Authentication Scheme
+This scheme uses the [OAuth2 Authorization Framework](http://tools.ietf.org/html/rfc6749) specified in [RFC 6749](http://tools.ietf.org/html/rfc6749) to get an authorization grant which is then used to access the API.
+
+The customizable options for this scheme are:
+
++ `oauth_resource_endpoint` is the oauth resource endpoint where the oauth framework resides. Value type is **string**. "/" is the **default**.
++ `scopes` represents the OAuth scopes which indicate a list of permissions. Value type is **array**. By **default** no scopes are specified.
```
kylef: Why are some of these options prefixed with `oauth_` and other are not (even though the descriptions both say they are OAuth specific).

> + `oauth_resource_endpoint` is the *oauth* resource endpoint where the oauth framework resides
> + `scopes` represents the *OAuth* scopes which indicate a list of permissions.

---

There is also inconsistency of the spelling of OAuth here, we have "OAuth{ in scopes and "oauth" in other places, can we make these consistent?

API Blueprint Specification.md
```diff
@@ -478,6 +490,232 @@ Name and description of the API
...
+```
+
+If the authentcation setting contains a list of scheme names after the `Authenticated` keyword, the transaction example only allows the specific authentication schemes denoted by the scheme names.
+
+```
++ Authenticated: oauth, token
+```
+
+```
++ Authenticated: oauth[read, email], token
```
honzajavorek: ah, so here is the example...

kylef: I haven't seen any documentation/mention about putting scopes in square brackets here. If this is supported, it should be mentioned/documented somewhere other than just this example.

API Blueprint Specification.md
```diff
@@ -478,6 +490,232 @@ Name and description of the API
...
+#### Resource Authenticated description
+Description of the resource authentication setting.
+
+If defined in a [Resource section](#def-resource-section), all the requests of the transaction examples under this resource will inherit the authentication setting unless specified otherwise.
+
+##### Example
+
+    # User [/user]
+
+    + Authenticated
```
kylef: Could we also place this under a resource group to indicate a whole resource group requires authentication?

API Blueprint Specification.md
pksunkara: It is. https://github.com/apiaryio/api-blueprint/pull/201/files#diff-c4fbef240fc70f2374b2048d50827f80R631

examples/14. Advanced Authentication.md
```diff
@@ -0,0 +1,64 @@
...
+
+        {
+          "username": "john",
+          "email": "john@appleseed.com"
+        }
+
+### Update a User Profile [POST]
+This endpoint allows the user to update an user profile only if the oauth scope of the access token is `user:write`.
+
++ Authenticated: oauth[`user:write`]
```
honzajavorek: Is this a sort of escaping? it wasn't mentioned anywhere in the spec above, not even in a side note. In case it's blueprint-wide way to escape things, I think scopes with `:` in them are so common that this needs to be at least in a note.

API Blueprint Specification.md
kylef: If this is oauth, it might be nice to describe this group or resource requires a particular oauth scope, along with how they are inherited (or overridden). (linking back to https://github.com/apiaryio/api-blueprint/pull/201/files?short_path=c4fbef2#r35647854)

```
+ Authenticated: oauth[user.read]
```

API Blueprint Specification.md
kylef: :+1: ah missed that one. :white_check_mark:

API Blueprint Specification.md
```diff
@@ -478,6 +490,232 @@ Name and description of the API
...
+#### OAuth2 Authentication Scheme
+This scheme uses the [OAuth2 Authorization Framework](http://tools.ietf.org/html/rfc6749) specified in [RFC 6749](http://tools.ietf.org/html/rfc6749) to get an authorization grant which is then used to access the API.
+
+The customizable options for this scheme are:
+
++ `oauth_resource_endpoint` is the oauth resource endpoint where the oauth framework resides. Value type is **string**. "/" is the **default**.
++ `scopes` represents the OAuth scopes which indicate a list of permissions. Value type is **array**. By **default** no scopes are specified.
++ `authorization_endpoint` is the authorization endpoint required for the authorization grant. Value type is **string**. "/authorize" is the **default**.
++ `access_token_endpoint` is the token endpoint used by client to obtain the access token. Value type is **string**. "/access_token" is the **default**.
++ `revocation_endpoint` is the revocation endpoint used by client to revoke a token. Value type is **string**. "/revoke" is the **default**.
```
kylef: I don't belive revocation is clearly define in RFC6746 and as such many services might not offer it. Therefore I don't think it should have a default value and when undefined it is not supported (opt-in).

API Blueprint Specification.md
```diff
@@ -478,6 +490,232 @@ Name and description of the API
...
++ `token_header_keyword` represents the string used instead of "Bearer" in "Authorization" header field according to [RFC 6750](http://tools.ietf.org/html/rfc6750#section-2.1). Value type is **string**. "Bearer" is the **default**.
++ `token_body_keyword` represents the parameter name used instead of "access_token" in the body of an API request according to [RFC 6750](http://tools.ietf.org/html/rfc6750#section-2.2). Value type is **string**. "access_token" is the **default**.
```
kylef: I've came across many APIs that don't support body and query authorisation tokens so I think we should have a way to state it's unsupported in a given API.

This may be unsupported due to security considerations:

> Because of the security weaknesses associated with the URI method (see Section 5), including the high likelihood that the URL containing the access token will be logged, it SHOULD NOT be used unless it is impossible to transport the access token in the "Authorization" request header field or the HTTP request entity-body. Resource servers MAY support this method.

API Blueprint Specification.md
```diff
@@ -478,6 +490,232 @@ Name and description of the API
...
+        + token_header_keyword: `token`
+        + scopes: read, write, email
+
+#### Bearer Authentication Scheme
+This scheme represents the method of sending a bearer token as specified in [RFC 6750](https://tools.ietf.org/html/rfc6750#section-2.1).
+
+The customizable options for this scheme are:
+
++ `keyword` represents the string used instead of "Bearer" in the "Authorization" header field according to [RFC 6750](https://tools.ietf.org/html/rfc6750#section-2.1). Value type is **string**. "Bearer" is the **default**.
++ `example_token` represents an example bearer token value. Value type is **string**. "example_token" is the **default**.
```
kylef: As per RFC6570 2.2 and 2.3, the token may be attached to form-encoded body parameter or URI Query Parameter. There isn't a way to customise these like you have added to the "OAuth2 Authentication Scheme".

API Blueprint Specification.md
```diff
@@ -478,6 +490,232 @@ Name and description of the API
...
+
+##### Example
+
+    + oauth (OAuth2 Authentication Scheme)
+        + oauth_resource_endpoint: `/oauth`
+        + access_token_endpoint: `/token`
+        + token_header_keyword: `token`
+        + scopes: read, write, email
+
+#### Bearer Authentication Scheme
```
kylef: I don't think I follow why this authentication scheme exists, the RFC that this scheme references is OAuth2 specific, this information is already included in the "OAuth2 Authentication Scheme" above.

API Blueprint Specification.md
honzajavorek: :+1: to prefixes... btw, @Almad had a good point about `oauth2_` prefix - are we sure we can't be bitten by using just `oauth_`?

honzajavorek:
:+1: to scopes being implemented, but I think it isn't documented perfectly and could use some more love.

API Blueprint Specification.md
kylef: Thinking about this more, I think we should allow the user to contain descriptions, perhaps something like:

```apib
+ scopes
    + `user:email` - Grants read access to a user’s email addresses.
```

Take a look at GitHub list of schemes which is more extensive and offers descriptions: https://developer.github.com/v3/oauth/#scopes

They also have inheritance of scopes: (user - Note that this scope includes user:email and user:follow.)

API Blueprint Specification.md
kylef: This is also unclear how this endpoint works, is it DELETE, POST? Does it revoke the current access token, all the user's access tokens?

API Blueprint Specification.md
honzajavorek: :+1: :+1: super useful, good idea, very wow such

API Blueprint Specification.md
zdne: While we shouldn't be disconnected with real world we should lead by example. @pksunkara can you please fix this?

API Blueprint Specification.md
zdne: @pksunkara were comments above addressed?

API Blueprint Specification.md
zdne: @pksunkara any comments on the previous remarks, I think @kylef is on spot

API Blueprint Specification.md
zdne: @pksunkara yes / no?

pksunkara: I can add that

API Blueprint Specification.md
pksunkara: If you look at the description of parameters, `N` is an example while `n` is a generic.

API Blueprint Specification.md
pksunkara: It is defined in [RFC 7009](https://tools.ietf.org/html/rfc7009) and is a POST work. This is to be used by the client by providing `client_id` and `client_secret`, not the user.

API Blueprint Specification.md
pksunkara: This exists because this is a popular auth scheme being used by a few closed APIs.

API Blueprint Specification.md
kylef: I wasn't aware of RFC7009, since this isn't actually part of the original RFC for OAuth2 (RFC6746) and instead added as a separate RFC a year later. There will be many implementations that don't offer this, or offer it in a different way. As such I still believe we should not have a default value and make this opt-in for the many APIs that do not support RFC7009.

For example, GitHub offer this in a different way in their OAuth2 implementation where they use DELETE. https://developer.github.com/v3/oauth_authorizations/#revoke-an-authorization-for-an-application

pksunkara:
Addressed reviews.

API Blueprint Specification.md
```diff
@@ -478,6 +491,257 @@ Name and description of the API
...
+- **Inherits from**: none
+
+#### Definition
+Defined by the `Authenticated` keyword in Markdown list entity followed by optional comma seperated list of scheme names which are prefixed by a colon.
+
+    + Authenticated: <list>
+
+If name of an authentication scheme dervied from `OAuth2 Authentication Scheme` is used, it can optionally have comma seperated list of OAuth scopes inside `[]` immediately following it. If two or more OAuth scopes are using in an authentication setting, then any of the scopes can be used to satisfy the authentication setting.
+
+> **NOTE:** Since each of the OAuth scopes is an [MSON Value][], they should be escaped by backticks if they contain any [MSON][] reserved characters
```
kylef: Missing full-stop at the end of the sentence.

examples/14. Advanced Authentication.md
kylef: This was amended above :white_check_mark:.

API Blueprint Specification.md
```diff
@@ -478,6 +491,257 @@ Name and description of the API
...
+---
+
+## Authenticated section
+- **Parent sections:** [Resource section](#def-resource-section), [Action section](#def-action-section), [Request section](#def-request-section)
+- **Nested sections:** none
+- **Markdown entity:** list
+- **Inherits from**: none
+
+#### Definition
+Defined by the `Authenticated` keyword in Markdown list entity followed by optional comma seperated list of scheme names which are prefixed by a colon.
```
kylef: separated*

API Blueprint Specification.md
```diff
@@ -478,6 +491,257 @@ Name and description of the API
...
+
+    + Authenticated: <list>
+
+If name of an authentication scheme dervied from `OAuth2 Authentication Scheme` is used, it can optionally have comma seperated list of OAuth scopes inside `[]` immediately following it. If two or more OAuth scopes are using in an authentication setting, then any of the scopes can be used to satisfy the authentication setting.
```
kylef: separated*

kylef: derived*

API Blueprint Specification.md
```diff
@@ -478,6 +491,257 @@ Name and description of the API
...
+
+The authentication setting for a [Request section](#def-request-section) denotes the authentication needed to be sent along with the request to get the [Response section](#def-response-section) of the respective transaction example.
+
+If the authentication setting contains only the `Authenticated` keyword, the transaction example allows all the authentication schemes described in the [Authentication Schemes section](#def-authentication-schemes-section).
+
+```
++ Authenticated
+```
+
+If the authentcation setting contains a list of scheme names after the `Authenticated` keyword, the transaction example only allows the specific authentication schemes denoted by the scheme names.
```
kylef: authentcation > authentication

zdne:
At its current form I have four major concerns: 

1. Syntax alignment with API Blueprint & MSON
2. Missing description of OAuth query attributes
3. Missing specification of OAuth query serialization
4. Missing the ability to describe BOTH authenticated AND unauthenticated situations

We will discuss these offline next week.

pksunkara:
> Missing the ability to describe BOTH authenticated AND unauthenticated situations

@zdne That is covered. Can you give me an example of what you are talking about?

zdne:
For the completeness: I have discussed my ideas on this proposal with Pavan offline. Here are drafts of some of the ideas – https://gist.github.com/zdne/b77576710b8a42106664

smizell:
@zdne @pksunkara There was a request tonight for [digest access auth](https://en.wikipedia.org/wiki/Digest_access_authentication) today. Will this proposal support that in the short or long term?

zdne:
@smizell not planned in the first go no. Only Basic, OAuth (2.0) and Token.

zdne:
## Status Update

This PR won't be accepted. Instead we will propose a new PR addressing concerns mentioned in https://github.com/apiaryio/api-blueprint/pull/201#issuecomment-131159425. 

Please stay tuned for a new proposal.

zdne:
This PR is superseded by the following API Blueprint RFC draft proposals: 

1. https://github.com/apiaryio/api-blueprint-rfcs/blob/master/draft/0002-authentication.md
1. https://github.com/apiaryio/api-blueprint-rfcs/pull/5
1. https://github.com/apiaryio/api-blueprint-rfcs/pull/6

Please follow and contribute on these proposals at the [API Blueprint RFC repository](https://github.com/apiaryio/api-blueprint-rfcs). Thanks you!

PR #213, opened 2015-06-15
Update 10. Data Structures.md

telaviv1:
Adding an example of a required parameter

### DIFF
--- a/examples/10. Data Structures.md
+++ b/examples/10. Data Structures.md
@@ -21,7 +21,7 @@ A coupon contains information about a percent-off or amount-off discount you mig
         The ID of the desired coupon.
 
 + Attributes (Coupon Base)
-    + id: 250FF (string)
+    + id: 250FF (string, required)
     + created: 1415203908 (number) - Time stamp
 
 ### Retrieve a Coupon [GET]


zdne:
Hey @telaviv1 thanks for contribution! I am definitely up for showing more of the MSON syntax in the advanced example. However these examples are trying to follow a continuity with what was in the previous installment. Here you are introducing `required` for something that was previously optional. Besides this is not a good practice in general (making something that used to be optional required) – because you are breaking the clients it may also confuse the readers who read the previous parts (no.9 and 8). 

At a minimum I would suggest to add an explanation at the top of this example. Or move it all the way to example no.8. What do you think?

telaviv1:
Add it to example number 8.  That's definitely the right place for it.

zdne:
@telaviv1 Makes sense. Note this needs updated of both 8, 9, and 10. Are you going to do this as a part of this PR?

pksunkara:
@telaviv1 Will you be working on this?

pksunkara:
Continued in #287

zdne:
Thanks @pksunkara !

PR #222, opened 2015-07-06
[Tutorial] Use API Blueprint for syntax

kylef:
This makes the tutorial much easier to follow when reading on GitHub.


honzajavorek:
:+1:

PR #230, opened 2015-07-17
Unmark Response as Abstract

ruslansagitov:
I guess, it’s a mistake.


kylef:
Thanks, that looks right. Just the Named, Asset and Payload section are under the abstract header at the top.

PR #243, opened 2015-08-06
Use APIB code blocks in the specification

kylef:
This changes every example that is using APIB to use the GFM apib highlighting.

    ```apib
    # Hello World
    ```


zdne:
:eyes: nice!

PR #278, opened 2015-11-04
Add JSON Schema example

danielgtaylor:
The example is pretty basic, but shows how a simple example could look and includes a list of strings.

### DIFF
--- a/examples/13. Named Endpoints.md
+++ b/examples/13. Named Endpoints.md
@@ -6,6 +6,7 @@ This API example demonstrates how to define a standalone endpoint with an identi
 ## API Blueprint
 + [Previous: Advanced Action](12.%20Advanced%20Action.md)
 + [This: Raw API Blueprint](https://raw.github.com/apiaryio/api-blueprint/master/examples/13.%20Named%20Endpoints.md)
++ [Next: JSON Schema](14.%20JSON%20Schema.md)
 
 # Group Quick start
 
--- a/examples/14. JSON Schema.md
+++ b/examples/14. JSON Schema.md
@@ -1,0 +1,93 @@
+FORMAT: 1A
+
+# JSON Schema
+Every request and response can have a schema. Below you will find examples using [JSON Schema](http://json-schema.org/) to describe the format of request and response body content.
+
+## API Blueprint
++ [Previous: Named Endpoints](13.%20Named%20Endpoints.md)
++ [This: Raw API Blueprint](https://raw.github.com/apiaryio/api-blueprint/master/examples/14.%20JSON%20Schema.md)
++ [Next: Advanced JSON Schema](15.%20Advanced%20JSON%20Schema.md)
+
+# Notes [/notes/{id}]
+
++ Parameters
+
+    + id: abc123 (required) - Unique identifier for a note
+
+## Get a note [GET]
+Gets a single note by its unique identifier.
+
++ Response 200 (application/json)
+
+    + Body
+
+            {
+                "id": "abc123",
+                "title": "This is a note",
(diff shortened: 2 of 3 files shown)


kylef:
Perhaps it would be nice to also include an example of an MSON described data structure with a JSON schema to emphasise you can override generated schemas for MSON structures.

examples/13. Named Endpoints.md
```diff
@@ -6,6 +6,7 @@ This API example demonstrates how to define a standalone endpoint with an identi
 ## API Blueprint
 + [Previous: Advanced Action](12.%20Advanced%20Action.md)
 + [This: Raw API Blueprint](https://raw.github.com/apiaryio/api-blueprint/master/examples/13.%20Named%20Endpoints.md)
++ [Next: JSON Schema](14.%20JSON%20Schema.md)
```
pksunkara: I don't know if JSON Schema has higher priority and should be listed a bit above in the examples directory. But that would mean editing a lot of examples. @zdne What is your directive on this? Should we just add the newer examples at the bottom?

examples/14. JSON Schema.md
```diff
@@ -0,0 +1,96 @@
...
+                      "items": {
+                          "type": "string"
+                      }
+                  }
+                }
+            }
+
++ Response 204
+
+## Delete a note [DELETE]
```
pksunkara: Do we need this action?

pksunkara:
> Perhaps it would be nice to also include an example of an MSON described data structure with a JSON schema to emphasise you can override generated schemas for MSON structures.

I think that should be another example on it's own along with the other Attributes examples.

Reviewed.

examples/14. JSON Schema.md
danielgtaylor: No, it was just added to round out the example. I will remove it.

danielgtaylor:
@kylef @pksunkara should I add another example with MSON or should that be a separate PR?

pksunkara:
@danielgtaylor You can add it to this PR.

examples/14. JSON Schema.md
```diff
@@ -0,0 +1,92 @@
...
+                "tags": [
+                    "todo",
+                    "work"
+                ]
+            }
+
+    + Schema
+
+            {
+                "type": "object",
```
kylef: Would additional unknown properties send in the request be allowed or rejected by the server?

Perhaps we should set `additionalProperties` in the schema to `false` to indicate that other unknown keys are not permitted in the request body.

kylef: Per [JSON Schema Validation 5.4.4 additionalProperties, properties and patternProperties](http://json-schema.org/latest/json-schema-validation.html#anchor64).

examples/14. JSON Schema.md
```diff
@@ -0,0 +1,92 @@
...
+                "properties": {
+                  "title": {
+                      "type": "string"
```
kylef: We're mixing 2-spaces and 4-spaces for indentation in these JSON structures. Can we align the indentation to be consistent across all JSON bodies.

Above `type` and `properties` is indented by 4 spaces, then `title` is only indented by `two` spaces and then `type` inside `title` is indented by another 4 spaces.

examples/13. Named Endpoints.md
zdne: @pksunkara I think in this case it is OK if this goes last (as an advanced topic).

pksunkara:
Just minor indentation issues.

danielgtaylor:
@kylef @pksunkara updated with your feedback. Please take another look!

PR #279, opened 2015-11-06
Create an advanced API Blueprint tutorial

danielgtaylor:
This (draft) creates a new advanced API Blueprint tutorial covering JSON Schema, request and response attributes, data structures and relation types.

I'm pretty unsure of the relation types section, so please have a look and give me any feedback. Once this is ready, I'm thinking of making available a final full example `.apib` file with these features and linking the tutorial to the advanced tutorial somehow.

Feedback will be much appreciated!

### DIFF
--- a/Advanced Tutorial.md
+++ b/Advanced Tutorial.md
@@ -1,0 +1,124 @@
+# Advanced API Blueprint Tutorial
+
+Welcome to the advanced API Blueprint tutorial! This tutorial will take you through advanced topics like JSON Schema, request and response attributes, data structures and relation types.
+
+This tutorial assumes that you have read the [API Blueprint Tutorial](./Tutorial.md).
+
+## JSON Schema
+
+Action request and response bodies can have associated schemas that describe the allowed structure of the body content. JSON bodies are typically described with [JSON Schema](http://json-schema.org/). Given a simple JSON response body we can describe the structure of the response with JSON Schema in a `+ Schema` section.
+
+The schema can describe the type of each member, which members are required, default values, and support a number of other advanced features. Below is an example, taken from the [Polls API](https://raw.github.com/apiaryio/api-blueprint/master/examples/Polls%20API.md) blueprint:
+
+```apib
+### Create a New Question [POST]
+You may create your own question using this action. It takes a JSON object containing a question and a collection of answers in the form of choices.
+
++ Request (application/json)
+
+    + Body
+
+            {
+              "question": "Favourite language?"
+              "choices": [
+                "Swift",
+                "Objective-C"
+              ]
+            }
+
+    + Schema
+
+            {
+              "$schema": "http://json-schema.org/draft-04/schema#",
+              "type": "object",
+              "properties": {
+                "question": {
+                  "type": "string
+                },
+                "choices": {
+                  "type": "array",
+                  "items": {
+                    "type": "string"
+                  },
+                  "minItems": 2
+                }
+              }
+            }
+```
+
+## Attributes
+
+Another way of describing examples and the structure of your request and response content is by using [MSON](https://github.com/apiaryio/mson#readme). MSON, like API Blueprint, allows you to use human-readable plain text to describe things rather than formats designed for computer parsing like JSON or YAML. Where API Blueprint allows you to describe your API, MSON allows you to describe data structure.
+
+MSON can be added to resources, actions, and individual requests or responses via an `+ Attributes` section.
+
+Creating a new question in the polls API can be modeled using MSON:
+
+```apib
+### Create a New Question [POST]
+You may create your own question using this action. It takes a JSON object containing a question and a collection of answers in the form of choices.
+
++ Request (application/json)
+
+    + Attributes
+
+        + question: Favourite Language? (required)
+        + choices: Swift, `Objective-C` (array, required)
+
+```
+
+When the above blueprint is parsed it will have a JSON body and JSON Schema example generated for it from the MSON attributes. Note, however, that the generated JSON Schema may differ from a hand-written one. In this example, the `minItems` will not be set. If you have such constraints you can override the generated schema by providing your own, in which case only the JSON body will be generated.
+
+*Note*: MSON is still considered a beta feature!
+
+## Data Structures
+
+Once you start using MSON, you may find yourself wanting to reuse certain commonly used or nested data structure components. This is possible with the `## Data Structures` section. Attributes sections can then reference the data structures defined in the Data Structures or other resource sections by name.
+
+For example, using the polls API question collection resource, we can split out the `Question` and `Choice` objects:
+
+```apib
+### List All Questions [GET]
++ Response 200 (application/json)
+
+    + Attributes (array[Question])
+
+## Data Structures
+
+### Question
++ question: Favourite programming language? (required)
++ published_at: `2014-11-11T08:40:51.620Z` (required)
++ url: /questions/1 (required)
++ choices (array[Choice], required)
+
+### Choice
++ choice: Javascript (required)
++ url: /questions/1/choices/1 (required)
++ votes: 2048 (number, required)
+
+```
+
+## Relation Types
+
+Actions in a blueprint can define a semantic domain-specific meaning by defining a relation type using a `+ Relation` section. This means that an action can have a specific meaning regardless of its URI or name, and allows clients to be built based on the domain of the API rather than specific URIs.
+
+For example, in the polls API:
+
+```apib
+## Question [/question/{id}]
+### View a Question Detail [GET]
++ Relation: self
+
+### Delete a Question [DELETE]
++ Relation: delete
+
+## Questions Collection [/questions]
+### List All Questions [GET]
++ Relation: self
+```
+
+A server or client implementation can now use this information to handle the specific API resource and action URIs. Please see the [Web Linking RFC 5988](https://tools.ietf.org/html/rfc5988) and the [IANA Link Relation Types](http://www.iana.org/assignments/link-relations/link-relations.xhtml) for more information.
+
+## Conclusion
+
+This tutorial has covered some advanced API Blueprint topics. For more in-depth information and other advanced topics, please see the [API Blueprint Specification](https://github.com/apiaryio/api-blueprint/blob/master/API%20Blueprint%20Specification.md).


Advanced Tutorial.md
```diff
@@ -0,0 +1,211 @@
...
+              {
+                "question": "Favourite programming language?",
+                "published_at": "2014-11-11T08:40:51.620Z",
+                "url": "/questions/1",
+                "choices": [
+                  {
+                    "choice": "Swift",
+                    "url": "/questions/1/choices/1",
+                    "votes": 2048
+                  }, {
```
pksunkara: Can we just have 2 choices here? We don't need as many. We should try to make the Tutorials as small as possible

danielgtaylor: This was copied directly from the first tutorial, but I can make it have fewer choices if it helps.

Advanced Tutorial.md
```diff
@@ -0,0 +1,211 @@
...
+
+        + (object)
+            + question: Favourite programming language? (required)
+            + published_at: `2014-11-11T08:40:51.620Z` (required)
+            + url: '/questions/1' (required)
+            + choices (array, required)
+                + (object)
+                    + choice: 'Javascript' (required)
+                    + url: '/questions/1/choices/1' (required)
+                    + votes: 2048 (number, required)
```
pksunkara: There is only 1 choice depicted here. Shouldn't a question have more than 1 choice?

Advanced Tutorial.md
```diff
@@ -0,0 +1,211 @@
...
+```apib
+## Question [/question/{id}]
+### View a Question Detail [GET]
++ Relation: question
+
+### Update a Question [PUT]
++ Relation: question_update
+
+### Delete a Question [DELETE]
++ Relation: question_delete
```
pksunkara: Do we need the `question` prefixes? It's not like there is a rule saying relations should be unique in the whole API Blueprint. What do you say @kylef, @zdne?

pksunkara:
Looks good. Haven't checked closely for grammar issues.

Advanced Tutorial.md
```diff
@@ -0,0 +1,211 @@
...
+                    "url": "/questions/1/choices/4",
+                    "votes": 256
+                  }
+                ]
+              }
+            ]
+
+    + Schema
+
+            {
```
kylef: This tutorial produces a large JSON Schema, I wonder if we would be better off describing a simpler structure, perhaps the structure used on creating a question instead:

```apib
+ Request (application/json)
    + Body

            {
              "question": "Favourite language?"
              "choices": [ "Swift", "Objective-C" ]
            }

    + Schema

            {
              "$schema": "http://json-schema.org/draft-04/schema#",
              "type": "object",
              "properties": {
                "question": { "type": "string }
                "choices": {
                  "type": "array",
                  "items": { "type": "string" },
                  "minItems": 2
                }
              }
            }
```

pksunkara: I agree with @kylef

Advanced Tutorial.md
```diff
@@ -0,0 +1,211 @@
...
+Welcome to the advanced API Blueprint tutorial! This tutorial will take you through advanced topics like JSON Schema, request and response attributes, data structures and relation types.
+
+This tutorial assumes that you have read the [API Blueprint Tutorial](./Tutorial.md).
+
+## JSON Schema
+
+Action request and response bodies can have associated schemas that describe the allowed structure of the body content. JSON bodies are typically described with [JSON Schema](http://json-schema.org/). Given a simple JSON response body we can describe the structure of the response with JSON Schema in a `+ Schema` section:
+
+```apib
+## Random Numbers [/random]
```
kylef: I think it would be nice to have a tutorial that keeps the same domain along all of the examples so it's easier to follow. Instead of jumping from a random number generator to the polls app, we could make both examples related to the polls app.

Advanced Tutorial.md
```diff
@@ -0,0 +1,211 @@
...
+                    + choice: 'Javascript' (required)
+                    + url: '/questions/1/choices/1' (required)
+                    + votes: 2048 (number, required)
+```
+
+Note: MSON is still considered a beta feature!
+
+## Data Structures
+
+Once you start using MSON, you may find yourself wanting to reuse certain commonly used or nested data structure components. This is possible with the `## Data Structures` section. Attributes sections can then reference the data structures defined in the Data Structures section by name.
```
kylef: I think it would be worth pointing out that you can reference both data structures and other resources by name.

Advanced Tutorial.md
```diff
@@ -0,0 +1,211 @@
...
+## Data Structures
+
+### Question
++ question: Favourite programming language? (required)
++ published_at: `2014-11-11T08:40:51.620Z` (required)
++ url: '/questions/1' (required)
++ choices (array[Choice], required)
+
+### Choice
++ choice: 'Javascript' (required)
```
kylef: Above you've placed strings directly and here wrapped them in quotes. Can we remove the quotes here to make this consistent with the other example `question` in Question above.

Advanced Tutorial.md
```diff
@@ -0,0 +1,211 @@
...
++ url: '/questions/1/choices/1' (required)
++ votes: 2048 (number, required)
+```
+
+## Relation Types
```
kylef: Seems to be a bug in GitHub's highlighting here when you view the diff. I wonder if the bug is caused from our `apib` lexer not completing correctly from the above code block? /cc @danielgtaylor

pksunkara: I think it's a markdown bug. It will be fixed if we add an empty line before the ending of above snippet with 3 backticks

Advanced Tutorial.md
```diff
@@ -0,0 +1,211 @@
...
+## Questions Collection [/questions]
+### List All Questions [GET]
++ Relation: questions
+```
+
+A server or client implementation can now use this information to handle the specific API resource and action URIs.
+
+## Conclusion
+
+This tutorial has covered some advanced API Blueprint topics. For more in-depth information and other advanced topics, please see the [API Blueprint Specification](https://github.com/apiaryio/api-blueprint/blob/master/API%20Blueprint%20Specification.md).
```
kylef: I think it would be nice to construct the full version of the above example(s) and link to it in the conclusion. So a user can view a complete blueprint the tutorial builds after the tutorial.

Advanced Tutorial.md
```diff
@@ -0,0 +1,211 @@
...
+
+For example, in the polls API:
+
+```apib
+## Question [/question/{id}]
+### View a Question Detail [GET]
++ Relation: question
+
+### Update a Question [PUT]
++ Relation: question_update
```
kylef: The Polls API doesn't allow you to update a question and I don't think it makes sense given our polls domain. Questions shouldn't be modifiable since users can vote on them.

I'd suggest we drop this action from the example.

Advanced Tutorial.md
kylef: I've called this relation `delete` in both the polls API (when you use hypermedia content types), and in the [polls-app](https://github.com/apiaryio/polls-app/blob/master/apiary.apib#L29)), we should certainly align these uses.

Is there any reason why you think `questions_delete` would be better over `delete`. I personally prefer the shorter variant so that a client can understand the semantic meaning of `delete` shared across multiple resources instead of a one-off relation which differs per resource.

Advanced Tutorial.md
```diff
@@ -0,0 +1,211 @@
...
+
+### Choice
++ choice: 'Javascript' (required)
++ url: '/questions/1/choices/1' (required)
++ votes: 2048 (number, required)
+```
+
+## Relation Types
+
+Actions in a blueprint can define a semantic domain-specific meaning by defining a relation type using a `+ Relation` section. This means that an action can have a specific meaning regardless of its URI or name, and allows clients to be built based on the domain of the API rather than specific URIs.
```
kylef: I'd love to see a link to the RFC defining [web linking](https://tools.ietf.org/html/rfc5988) either here or in the conclusion for this section of the tutorial. This allows a user to go off and read more.

Advanced Tutorial.md
pksunkara: What would do if there are 2 delete methods in the API? 1 for questions, 1 for choices.

Advanced Tutorial.md
```diff
@@ -0,0 +1,211 @@
...
+
+For example, in the polls API:
+
+```apib
+## Question [/question/{id}]
+### View a Question Detail [GET]
++ Relation: question
```
kylef: I think a relation to the same resource should be [`self`](http://www.iana.org/assignments/link-relations/link-relations.xhtml).

The [polls app](https://github.com/apiaryio/polls-app/blob/master/apiary.apib#L60) is also incorrect, this also needs to be changed but depends on [Draft API Blueprint RFC for affordances](https://github.com/apiaryio/api-blueprint-rfcs/pull/2).

Advanced Tutorial.md
danielgtaylor: It is an array of choices with a single sample. It's not entirely clear to me how to add more samples when you have an array of objects. Do you know how?

Advanced Tutorial.md
danielgtaylor: @kylef I will rename them to match the polls app, but how do you link between resources in your approach?

danielgtaylor:
@kylef @pksunkara I think I have addressed all of the feedback. Please take another look!

I'll also look into the highlighting bugs soon, but that will be a different task.

Advanced Tutorial.md
```diff
@@ -0,0 +1,119 @@
...
+
+Welcome to the advanced API Blueprint tutorial! This tutorial will take you through advanced topics like JSON Schema, request and response attributes, data structures and relation types.
+
+This tutorial assumes that you have read the [API Blueprint Tutorial](./Tutorial.md).
+
+## JSON Schema
+
+Action request and response bodies can have associated schemas that describe the allowed structure of the body content. JSON bodies are typically described with [JSON Schema](http://json-schema.org/). Given a simple JSON response body we can describe the structure of the response with JSON Schema in a `+ Schema` section.
+
+The schema can describe the type of each member, which members are required, default values, and support a number of other advanced features. Below is an example, taken from the API Blueprint tutorial blueprint:
```
pksunkara: The below example is not taken from API Blueprint tutorial.

Advanced Tutorial.md
```diff
@@ -0,0 +1,119 @@
...
+            }
+```
+
+## Attributes
+
+Another way of describing examples and the structure of your request and response content is by using [MSON](https://github.com/apiaryio/mson#readme). MSON, like API Blueprint, allows you to use human-readable plain text to describe things rather than formats designed for computer parsing like JSON or YAML. Where API Blueprint allows you to describe your API, MSON allows you to describe data structure.
+
+MSON can be added to resources, actions, and individual requests or responses via an `+ Attributes` section.
+
+The polls API question resource can be modeled using MSON:
```
pksunkara: The below example is not a polls API question resource.

danielgtaylor:
@pksunkara addressed feedback. Thanks!

Advanced Tutorial.md
```diff
@@ -0,0 +1,123 @@
...
+
+When the above blueprint is parsed it will have a JSON body and JSON Schema example generated for it from the MSON attributes.
+
+*Note*: MSON is still considered a beta feature!
+
+## Data Structures
+
+Once you start using MSON, you may find yourself wanting to reuse certain commonly used or nested data structure components. This is possible with the `## Data Structures` section. Attributes sections can then reference the data structures defined in the Data Structures or other resource sections by name.
+
+For example, using the polls API question resource, we can split out the `Question` and `Choice` objects:
```
pksunkara: Isn't `List All Questions` part of `Question Collection` resource?

Advanced Tutorial.md
```diff
@@ -0,0 +1,123 @@
...
++ question: Favourite programming language? (required)
++ published_at: `2014-11-11T08:40:51.620Z` (required)
++ url: /questions/1 (required)
++ choices (array[Choice], required)
+
+### Choice
++ choice: Javascript (required)
++ url: /questions/1/choices/1 (required)
++ votes: 2048 (number, required)
+```
```
pksunkara: Add an empty line before this line and the syntax highlighting should be fixed.

Advanced Tutorial.md
```diff
@@ -0,0 +1,123 @@
...
++ Request (application/json)
+
+    + Attributes
+
+        + question: Favourite Language? (required)
+        + choices: Swift, `Objective-C` (array, required)
+
+```
+
+When the above blueprint is parsed it will have a JSON body and JSON Schema example generated for it from the MSON attributes.
```
kylef: Perhaps it might be worth noting a user can override the schema, and the generated schema won't be the same as above as it won't include `minItems`.

danielgtaylor:
More feedback addressed. I can't seem to fix the syntax highlighting - it will need to be done along with a few other minor fixes in the syntax highlighter package. Will try and get to that soon, as it's quite annoying.

zdne:
👍

PR #282, opened 2015-11-11
Add API Blueprint Slack badge

zdne:
This adds a badge and the link to the API Blueprint Slack sign-up page (app).

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@
 # API Blueprint
 ### API Design for Humans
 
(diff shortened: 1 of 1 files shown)


Baggz:
:dancer:

Baggz:
BTW, this is how the invitation looks like.

image

zdne:
@Baggz yes, we are aware of it. We will leave it as is right now since we are running on a free dyno and Lada would like to know if we will run over its limits.

PR #298, opened 2016-01-04
Add npm package of the documentation

w-vi:


zdne:
@kylef @w-vi why are we doing this?

pksunkara:
I agree with Z. Why are we doing this?

kylef:
Yes, this is an approach I don't want to go down. I believe with some discussion we can proceed by using a git submodule now so I'm going to close this.

w-vi:
It really sucks, so maybe next we will be able to persuade the rest that submodules are the best option

zdne:
@w-vi @kylef @pksunkara I did not say this is good or bad. Rather, I was missing any explanation of this PR.

PR #299, opened 2016-01-07
Remove Paramerer enum value descriptions from the specification

kylef:
These are not supported in any parser, Refract nor the API Blueprint AST and therefore should be removed from the specification.

This will be later supported once RFC3 has been implemented (https://github.com/apiaryio/api-blueprint-rfcs/pull/3).


zdne:
Just for the record. This means following wont't be supported for the time being:

```apib
# API 
## GET /{i}
- parameters
    - i: 2 (enum)
        - Members
           - 1 - Hello World!
           - 2
- response 200
```

PR #301, opened 2016-01-18
Change API Blueprint logo

pksunkara:


kylef:
Could we also update the image reference from the README so it's relative instead of referencing the "master" branch so this can be previewed?

```diff
- ![logo](https://raw.github.com/apiaryio/api-blueprint/master/assets/logo_apiblueprint.png)
+ ![API Blueprint Logo](assets/logo_apiblueprint.png)
```

pksunkara:
@kylef It won't work all the time. It will work only in github. We should encourage direct links instead of github relative links which only work in github. I think using `raw.github.com` is better.

kylef:
> It won't work all the time. It will work only in github

Can you elaborate on when it won't work? I can't think of any cases this wouldn't work. Unless you only had the README.md file and not the image.

> We should encourage direct links instead of github relative links which only work in github.

The biggest problem with this is that it references master which means all historic versions of the README will not be correct since "master" is a moving target.

pksunkara:
Updated.

PR #324, opened 2016-04-28
Wrap lines, use semantic linebreaks

jean:
I noticed that the `### Create a New Question [POST]` example
rendered in a box with horisontal scrollbar, with most of the
description scrolling off to the right.

When going to fix this, I noticed that descriptive text wasn't wrapped.
Wrap lines, it's more suited to a plaintext format like markdown.

Semantic linebreaks:
  http://rhodesmill.org/brandon/2012/one-sentence-per-line/

### DIFF
--- a/Advanced Tutorial.md
+++ b/Advanced Tutorial.md
@@ -12,7 +12,8 @@ The schema can describe the type of each member, which members are required, def
 
 ```apib
 ### Create a New Question [POST]
-You may create your own question using this action. It takes a JSON object containing a question and a collection of answers in the form of choices.
+You may create your own question using this action. It takes a JSON object
+containing a question and a collection of answers in the form of choices.
 
 + Request (application/json)
 
@@ -56,7 +57,8 @@ Creating a new question in the polls API can be modeled using MSON:
 
 ```apib
 ### Create a New Question [POST]
-You may create your own question using this action. It takes a JSON object containing a question and a collection of answers in the form of choices.
+You may create your own question using this action. It takes a JSON object
+containing a question and a collection of answers in the form of choices.
 
 + Request (application/json)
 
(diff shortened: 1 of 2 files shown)


pksunkara:
Can you please squash this into a single commit?

jean:
I'll do so and let you know :-)
I'm going through the spec now.

pksunkara:
@jean Actually, after giving it a thought, I realised we don't need to wrap everything. Can you only wrap the examples which are nested in code blocks? And don't do the semantic line breaks. Just wrapping them is enough.

jean:
@pksunkara well, I realise it's to some extent a matter of preference. For rendering, paragraphs don't *need* to be wrapped. But markdown uses a double-linebreak as paragraph separator so that paragraphs *can* be wrapped. As Gruber writes:

> Readability, however, is emphasized above all else. A Markdown-formatted document should be publishable as-is, as plain text, without looking like it’s been marked up with tags or formatting instructions.

For me, this includes wrapping lines. A lot of source code, such as C or javascript, doesn't need wrapping either. But yet we do, for readability and better diffs. 

Here's a markdown style guide that raises the same points (both pro and con):
  http://www.cirosantilli.com/markdown-style-guide/#line-wrapping

So with my pull request, I'd like to advocate for wrapping the markdown source at around 80 characters.

pksunkara:
@jean I understand. But can we wrap the examples first in this PR? You can make another PR to wrap all other markdown.

PS: No semantic line breaks though.

jean:
@pksunkara OK! I'll prepare a new PR.

jean:
@pksunkara this PR now only wraps descriptions in examples.

pksunkara:
Thanks for the PR! 😄

jean:
Thanks for the tools :)

PR #393, opened 2017-04-30
add common data description

goganchic:
Implements https://github.com/apiaryio/drafter/issues/467


w-vi:
Shouldn't it be named `Common Response` rather then data if it holds responses?

goganchic:
I think this section in future will contain common headers also. That is why I've called it `Common Data`

abrahamcuenca:
This PR is quite stale. Should we close it?

abtris:
Good point, I will close it.

